PR #20, opened 2021-02-22
add linkcheck step and fail build on warning

### LINKED ISSUE #14: Add link checking and linting to actions build
As discussed on slack and zoom, I'm going to try adding `make linkcheck` and RST linting to the build process so that the kind of issues being addressed in https://github.com/natcap/invest.users-guide/pull/11 don't accumulate again.

emlys:
Fixes #14 
- Run `make linkcheck`. This checks that every external link in our docs is up-to-date. Fails if there are any errors, redirects are ok.
- Add `-W` flag to `make html` command that raises warnings to errors. Pretty much all sphinx syntax errors are only warnings in the build so this is necessary to fail the build.

I looked into some RST linting options and it seems best to just go by the sphinx build warnings.
- `rstcheck` hasn't had a new release since 2018 and it seemed to display incomplete errors
- `restructuredtext-lint` by default only supports the standard docutils directives, not sphinx ones. You have to custom register the sphinx directives you use, and it's unclear if that's possible for all of them
- `doc8` uses `restructuredtext-lint` and handles the sphinx issues, but it's particular about line length and carriage returns


source/croppollination.rst
```diff
@@ -312,7 +312,7 @@ Klein, AM, BE Vaissiere, JH Cane, I. Steffan-Dewenter, SA Cunningham, C. Kremen,
 
 Kremen, C., NM Williams, RL Bugg, JP Fay, and RW Thorp. 2004. The area requirements of an ecosystem service: crop pollination by native bee communities in California. Ecology Letters 7: 1109-1119.
 
-Lonsdorf, E., C. Kremen, T. Ricketts, R. Winfree, N. Williams, and SS Greenleaf. 2009. Modelling pollination services across agricultural landscapes.  Annals of Botany 1: 12. online [https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2701767/pdf/mcp069.pdf]
+Lonsdorf, E., C. Kremen, T. Ricketts, R. Winfree, N. Williams, and SS Greenleaf. 2009. Modelling pollination services across agricultural landscapes.  Annals of Botany 1: 12. online [htps://pubmed.ncbi.nlm.nih.gov/19324897/]
```
dcdenu4, line 315: Short a `t` here? `htps`

dcdenu4: Should this have been caught by linkcheck?

dcdenu4 (changes requested):
Hey @emlys ,

Link checking seems way overdue for stable Users Guide builds. I had one question and minor fix related to an updated link. 

I was also wondering if these changes still play well with InVESTS makefile `make userguide` command. Thanks!

source/croppollination.rst
emlys: whoops! Looks like htps redirects to https, so it passed, but I'll fix that

emlys:
@dcdenu4 thanks! Fixed that one url. It should still work with `make userguide`, the `PANDOC` isn't used anywhere and I think we would still want it to fail on sphinx warnings.

dcdenu4 (approved):
Thanks!

PR #21, opened 2021-02-22
Remove UI screenshots

### LINKED ISSUE #15: Remove screenshots of UI and spreadsheets
Some model documentation includes screenshots of the InVEST UI and of example spreadsheets which are prone to becoming outdated. I don't think we need to have screenshots of the UI at all. Example spreadsheets should be stored as CSVs. It would be extra cool if we could link to the external sample data rather than duplicating it here.
Noticed this in the following UG pages:

- [ ] fisheries
- [ ] habitat quality
- [ ] hra
- [ ] invest api
- [ ] marine fish
- [ ] ndr
- [ ] wave energy

emlys:
Fixes #15 
- Remove all screenshots of UI
- Remove unused files and add a step to the actions workflow that checks for unused files
- Replace sample tables with tables from the invest-sample-data repo
- Added `make sampledata` command that checks out the sample data
    - For a few sample data tables, makes a modified version (shorten long tables, replace \ with \\)


source/reservoirhydropowerproduction.rst
```diff
@@ -389,31 +389,6 @@ The simplest categorization of LULCs on the landscape involves delineation by la
...
-  11     Cropland (row Crops)
-  12     Bare Ground
-  13     Urban and Built-Up
-  14     Wetland
-  15     Mixed evergreen
-  16     Mixed Forest
-  17     Orchards/Vineyards
-  18     Pasture
-  ====== ===========================
-
```
emlys, line 416: what do you think about these? the same LULC section with table is repeated in a few models. I took this one out because it seemed unnecessary and potentially confusing (this table is not actually an input to anything)... but maybe it is useful to explain the concept of landcover codes.

.github/workflows/build-sphinx.yml
```diff
@@ -32,6 +32,30 @@ jobs:
             - name: Install python dependencies
               run: python -m pip install -r requirements.txt
 
+            - name: Check for unused files
```
dcdenu4, line 35: I tried thinking of a situation where we wouldn't mind having an unused file hanging around and I couldn't really come up with anything. I think it's fine to be strict and clean about this until a situation arises that makes us think differently.

Makefile
```diff
@@ -33,3 +33,33 @@ linkcheck: $(SOURCEDIR)
...
+	git -C $(GIT_SAMPLE_DATA_REPO_PATH) lfs fetch
+	git -C $(GIT_SAMPLE_DATA_REPO_PATH) checkout $(GIT_SAMPLE_DATA_REPO_REV)
+
+	# modifications to certain sample data files so they display nicely
+	# single backslashes don't get rendered in the csv-table, replace with double backslashes
+	sed 's/\\/\\\\/g' invest-sample-data/HabitatRiskAssess/Input/habitat_stressor_info.csv > invest-sample-data/HabitatRiskAssess/Input/habitat_stressor_info_modified.csv
+
+	# selections of tables that are too long to display in full
+	head -n1 invest-sample-data/pollination/landcover_biophysical_table.csv > invest-sample-data/pollination/landcover_biophysical_table_modified.csv 
+	tail -n3 invest-sample-data/pollination/landcover_biophysical_table.csv >> invest-sample-data/pollination/landcover_biophysical_table_modified.csv
```
dcdenu4, line 53: nifty!

Makefile
```diff
@@ -33,3 +33,33 @@ linkcheck: $(SOURCEDIR)
...
+	tail -n4 invest-sample-data/Carbon/carbon_pools_willamette.csv >> invest-sample-data/Carbon/carbon_pools_willamette_modified.csv
+
+	head -n11 invest-sample-data/WaveEnergy/input/Machine_Pelamis_Performance.csv > invest-sample-data/WaveEnergy/input/Machine_Pelamis_Performance_modified.csv
+
+	head -n4 invest-sample-data/WindEnergy/input/NE_sub_pts.csv > invest-sample-data/WindEnergy/input/NE_sub_pts_modified.csv
+
+
+
+
+
```
dcdenu4, line 65: Some dead space here.

source/carbon_edge.rst
```diff
@@ -196,34 +196,18 @@ The biophysical table columns must be named as follows:
...
+          :header-rows: 1
+          :widths: auto
+
+   *Sources for c_above data*: 
+
+   Closed shrublands, open shrublands, grasslands: Miranda, Sabrina do Couto, et al. Regional variations in biomass distribution in Brazilian savanna woodland. Biotropica 46.2 (2014): 125-138. 
+
+   Woody savannas, savanna: Saatchi, Susan S., et al. Distribution of aboveground live biomass in the Amazon basin. Global Change Biology 13.4 (2007): 816-837. 
+
+   Croplands: Fearnside, Philip M. Greenhouse gases from deforestation in Brazilian Amazonia: net committed emissions. Climatic Change 35.3 (1997): 321-360.
```
dcdenu4, line 210: I notice that we were super-scripting values in the table to indicate which source they came from... This is probably information we should try to replicate if we can. Maybe there's some clean way to say which rows from the table go with which source for the `c_above` values?

dcdenu4: I also noticed this table gets a `Table 4` designation, but it's the first table in this doc...

source/carbonstorage.rst
```diff
@@ -87,17 +87,10 @@ This section outlines the specific data used by the model. See the Appendix for
...
-  1      Forest              140     70      35     12
-  2      Coffee              65      40      25     6
-  3      Pasture/grass       15      35      30     4
-  4      Shrub/undergrowth   30      30      30     13
-  5      Open/urban          5       5       15     2
-  ====== ================== ======= ======= ====== ======
+  .. csv-table:: **Example Carbon Pools Table**
+       :file: ../invest-sample-data/Carbon/carbon_pools_willamette_modified.csv
+       :header-rows: 1
+       :widths: auto
```
dcdenu4, line 93: I wonder if we can still include the table description here... not that it really is providing much context you couldn't get from looking at the table values.

dcdenu4: Like the carbon edge table, this one is labeled `Table 5`. I guess it's just incrementing on the whole doc...

source/coastal_vulnerability.rst
```diff
@@ -61,7 +61,7 @@ Table 4.1
 """""""""
 
 .. csv-table:: **Example Ranking System**
-      :file: ./coastal_vulnerability/cv_table_41.csv
+      :file: ./coastal_vulnerability/ranking_system_table.csv
```
dcdenu4, line 64: This looks great. I am seeing `Table 8` instead of the before `Table 4`. The `Table 4.1: List ...` then doesn't line up... Not sure what's throwing off table numbers.

source/coastal_vulnerability.rst
```diff
@@ -340,8 +340,9 @@ The model uses a graphical interface to input all required and optional data. Be
          Sample: InVEST\CoastalVulnerability\GrandBahama_Habitats\Natural_Habitats.csv
 
    .. csv-table:: **Example Habitat Table**
-      :file: ./coastal_vulnerability/NaturalHabitat_WCVI.csv
+      :file: ../invest-sample-data/CoastalVulnerability/GrandBahama_Habitats/Natural_Habitats.csv
```
dcdenu4, line 343: Again just mentioning table number differences.

source/croppollination.rst
```diff
@@ -159,13 +159,16 @@ This section outlines the specific data used by the model. See the Appendix for
...
+
  * *nesting_[SUBSTRATE]_availability_index*: Relative index of the availability of the given nesting type within each LULC type, on a floating point scale of 0-1.  The *SUBSTRATE* name must exactly match a substrate given in the Guild Table.
 
  * *floral_resources_[SEASON]_index*: Relative abundance (floating point value 0-1) of flowers in each LULC class for the given season. There are two aspects to consider when estimating the relative floral abundance of each LULC class: % floral abundance or % floral coverage, as well as the duration of flowering during each season. For example, a land cover type comprised 100% of a mass flowering crop that flowers the entire season with an abundance cover of 80% would be given a suitability value of 0.80. A land cover type that flowers only half of the season at 80% floral coverage would be given a floral suitability value of 0.40.  The *SEASON* name must exactly match a season given in the Guild Table.
 
 .. csv-table:: **Example Biophysical Table**
-       :file: ./croppollination/landcover_biophysical_table_sample.csv
+       :file: ../invest-sample-data/pollination/landcover_biophysical_table_modified.csv
```
dcdenu4, line 169: This table has `Table 2` instead of original `Table 1`. It also has a column `descript`. Should that really be `description` and changed in the sample data? I'm assuming the model doesn't care and never looks for such a column, but it does look a little sloppy in the Users' Guide.

source/croppollination.rst
```diff
@@ -187,11 +190,9 @@ This section outlines the specific data used by the model. See the Appendix for
  *Example:* A hypothetical Guilds Table with two species. There are two main SUBSTRATEs, "cavity" and "ground." Species "Apis" uses both cavity and ground nesting types, and species "Bombus" only uses cavity nests. There are two SEASONs, "spring" and "summer".  Typical flight distances, specified in meters (alpha), vary widely between species. The relative_abundance of Bombus is higher than Apis, indicating that there are more Bombus pollinators than Apis.
 
 .. csv-table:: **Example Guild Table**
-       :file: ./croppollination/guild_table_sample.csv
+       :file: ../invest-sample-data/pollination/guild_table.csv
```
dcdenu4, line 193: This is showing as `Table 3` instead of original `Table 2`.

source/globio.rst
```diff
@@ -513,37 +499,10 @@ lucode globio_lucode
...
-msa_lu        Land Cover Class 3           0.5   0.03
-msa_lu        Land Cover Class 4           0.2   0.04
-msa_lu        Land Cover Class 5           0.7   0.05
-msa_lu        Land Cover Class 6           0.1   0.07
-msa_lu        Land Cover Class 7           0.5   0.06
-msa_lu        Land Cover Class 8           0.3   0.12
-msa_lu        Land Cover Class 9           0.1   0.08
-msa_lu        Land Cover Class 10          0.05  na
-============= ================ =========== ===== =====
+    .. csv-table:: **Example**
```
dcdenu4, line 502: This table listed as `Table 10` as opposed to be unlisted originally.

source/habitat_risk_assessment.rst
```diff
@@ -425,9 +423,13 @@ Criteria Scores CSV
 
 The Criteria Scores CSV (or Excel) file will provide all the criteria information for the run of the Habitat and Species Risk Assessment model. This file contains information about the effect of each stressor on each habitat (i.e. the exposure and consequence scores) for the habitats and stressors in your analysis. A template for the criteria CSV file can be found in the sample data folder. Users should feel free to add or remove specific criteria, and fill in ratings on a scale of 1 to 3, or 1 to any other value, so long as the scale is the same for all criteria
 
-.. figure:: habitat_risk_assessment/criteria_csv.PNG
+.. This image is of the same table provided in sample data.
+   It was too large and complex to format well as a csv-table so I'm leaving it as a figure.
```
dcdenu4, line 427: Yup, makes sense.

source/ndr.rst
```diff
@@ -238,9 +238,10 @@ You may choose to run the model with either Nitrogen or Phosphorus or both at th
   Example biophysical table (only to be used as an example, your LULC classes and corresponding values will be different):
 
   .. csv-table::
-    :file: ndr/ndr_biophysical_table_example.csv
+    :file: ../invest-sample-data/NDR/biophysical_table_gura.csv
```
dcdenu4, line 241: Do you think that showing a table with extra columns in this case is a little confusing? I also notice that the sample data uses `eff_p` as opposed to `eff_n`. I feel like in this case the sample data might not be robust to helping explain the capabilities of the model? 

I also noticed that some of the descriptions above mention things like:
>For example, high values (0.6 to 0.8) may be assigned to all natural vegetation types (such as forests, natural pastures, wetlands, or prairie)

These don't necessarily jive (jibe?) with what we have in the sample data.

source/recreation.rst
```diff
@@ -114,8 +114,9 @@ The following outlines the options presented to the user after opening the InVES
     .. warning:: All GIS data referenced in the Predictor Table must be in the same projected coordinate system as the Area of Interest shapefile. All distance, length, and area calculations use the same units as the AOI coordinate system.
 
     .. csv-table:: **Example Predictor Table**
-       :file: ./recreation/predictors.csv
+       :file: ../invest-sample-data/recreation/predictors.csv
```
dcdenu4, line 117: Just noting that the table is `Table 6` as opposed to the original `Table 3`.

source/reservoirhydropowerproduction.rst
dcdenu4: Yea, I ultimately think it's useful in this context since this is an *appendix* zone calling out useful data sources nuggets. 

However, I wish that we had a User's Guide section called **InVEST Data Resources** (or something), where we collated information on inputs for ALL models that we could point users to and update as we learned more (like from forum conversations).

source/sdr.rst
```diff
@@ -471,32 +469,6 @@ The simplest categorization of LULCs on the landscape involves delineation by la
 
 A slightly more sophisticated LULC classification involves breaking relevant LULC types into more meaningful types. For example, agricultural land classes could be broken up into different crop types or forest could be broken up into specific species. The categorization of land use types depends on the model and how much data is available for each of the land types. You should only break up a land use type if it will provide more accuracy in modeling. For instance, only break up 'crops' into different crop types if you have information on the difference in USLE C values between crops.
 
-*Sample Land Use/Land Cover Table*
-
-  ====== ===========================
-  lucode Land Use/Land Cover
-  ====== ===========================
-  1      Evergreen Needleleaf Forest
```
dcdenu4, line 479: Ah, interesting. This is annoyingly redundant. I wonder, outside of it's own section like I mentioned above, if this is a case where having a CSV stored in the UG repo makes sense, so it can be re-used in places like this? 

Unless we decide this isn't helpful at all.

source/wave_energy.rst
```diff
@@ -80,11 +78,11 @@ The ability of a WEC device to harvest wave energy can be expressed by wave ener
 
 .. Table 3.2 shows an example of wave energy absorption performances in each seastate bin for Pelamis.
  
-.. figure:: ./wave_energy/table_energyabsorption.png
-   :align: center
-   :figwidth: 500px
+.. csv-table:: **Wave energy absorption performance (kW) for Pelamis**
```
dcdenu4, line 81: A few things here:
- I noticed the `Figure` number changed. I also noticed a reference to `Table 3.1` that might have been that commented out `image002`? 
- I wonder if keeping this `png` table makes sense here, but later on for the **Data Needs**, it makes sense to show what that table really looks like as a data input?

source/wave_energy.rst
```diff
@@ -148,19 +146,21 @@ First we describe required inputs.  The required inputs are the minimum data nee
 	File type: *.csv
 	Sample data set:  \InVEST\WaveEnergy\Input\Machine_Pelamis_Performance
 
-.. figure:: ./wave_energy/table_pelamisperformance.png
-   :align: center
-   :figwidth: 500px
+.. csv-table::
```
dcdenu4, line 149: The indentation between the numbered inputs and the associated tables doesn't look great, is it possible to bump out the tables too? I also wonder if using your modifying approach might be good here to truncate some of these larger tables where more rows / cols doesn't really provide a lot of benefit?

dcdenu4 (changes requested):
Hey @emlys,

This is an interesting challenge! Overall I like the idea that UG model examples should be based on the sample data and having one source of truth for that makes sense. However, I go back and forth on how to think of the UG. Is it more of a publication or more of a detailed how-to guide? If the model developers think it should read more like a publication then maybe some of those tables / figures that don't really represent the sample data, but do provide a more simple example of understanding make sense. In some cases where the UG already is trying to reflect the sample data, then I think it makes total sense to replace those with the direct sample data connection. 

Some other things I noticed that should be looked at more closely:
- Table and figure numbers have changed significantly. I don't think a lot of the models reference these numbers, but some of them do and no longer reads well.
- Do we need to make changes to how InVEST `make userguide` operates as well?
- I'm assuming all those `crop_production` files that were deleted were unused? Are those files represented in the sample data? I'm just wondering if those would / were useful to someone and they should live elsewhere.
- I noticed that the `Fisheries` table now had some vertical table lines that make an empty column

Happy to chat about this more!

.github/workflows/build-sphinx.yml
emlys: Agree, they were really piling up and I was using this script to find files to delete anyway, so I figured I'd add it to the workflow

source/carbon_edge.rst
emlys: I was sort of trying to replicate it by putting `*Sources for c_above data*: ` and then the row descriptions but it's not as nice as the superscripts. Side note, it would be great to have this source information for all sample data! I'm not always sure if they're real data or just made up.

source/carbonstorage.rst
emlys: I put the description back and changed it to make sense with the data

source/wave_energy.rst
emlys: The `Table 3.1` reference seems to be about the seastate table. I just change it to say "The table below". And I'll put the png back here.

source/wave_energy.rst
emlys: I had tried to line up the tables within the numbered items and had no luck but someone on stackoverflow just pointed me in the right direction so it's fixed now!

dcdenu4:
Hey @emlys , is this PR back to me? No rush, I just wanted to make sure I wasn't holding anything up. Happy to chat about this more as well!

emlys:
@dcdenu4 Sorry, I've been totally focused on the stormwater model! I addressed a bunch of the table issues but I don't think I quite got to all of your questions. Is it alright to leave this for a couple more days?

source/reservoirhydropowerproduction.rst
emlys: Yea that would be great! Made an issue for it: #23

source/croppollination.rst
emlys: Hmm yea. May as well change that in the sample data.

emlys:
> If the model developers think it should read more like a publication then maybe some of those tables / figures that don't really represent the sample data, but do provide a more simple example of understanding make sense. In some cases where the UG already is trying to reflect the sample data, then I think it makes total sense to replace those with the direct sample data connection.

That's a good point, after going back thru it I agree some make more sense as a simplified example that's specific to the user's guide. I put back the original tables in some cases that you pointed out above. Let me know what you think

> Table and figure numbers have changed significantly. I don't think a lot of the models reference these numbers, but some of them do and no longer reads well.

Hmm, yea. Realizing I really don't know how the table numbering works. After a bit of googling I think it's a problem to put off until really necessary. Sphinx wants to number them relative to the entire user's guide, not by page. For now I've just removed the table headers that trigger auto-numbering. We can reference them by the `:label:` instead.

> Do we need to make changes to how InVEST `make userguide` operates as well?

I think it's unaffected

> I'm assuming all those `crop_production` files that were deleted were unused? Are those files represented in the sample data? I'm just wondering if those would / were useful to someone and they should live elsewhere.

There was some overlap with the sample data but not entirely. I guess if someone needs them in the future, they'll be there in the git history?

> I noticed that the `Fisheries` table now had some vertical table lines that make an empty column

I added the vertical lines to all the tables since some were hard to read without it. That empty column is in the sample data and it does look bad. I'll remove it from that table.

emlys:
Also addressing https://github.com/natcap/invest.users-guide/issues/22 by
- configuring the user-agent
- setting `linkcheck` to ignore a troublesome link
- adding a 2 minute timeout to the actions step to prevent the issue where it ran for 6 hours

source/coastal_vulnerability.rst
```diff
@@ -53,18 +53,12 @@ The primary output of the model is a geospatial dataset (points) that are plotte
...
+The model computes the coastal exposure index by combining the ranks of up to seven biological and physical variables at each shoreline point. Ranks vary from **very low exposure (rank=1)** to **very high exposure (rank=5)**, based on a mixture of user- and model-defined criteria. This ranking system is based on methods proposed by Gornitz et al. (1990) and Hammar-Klose and Thieler (2001).
 
-.. _table-41:
-
-Table 4.1
-"""""""""
-
-.. csv-table:: **Example Ranking System**
-      :file: ./coastal_vulnerability/cv_table_41.csv
+.. csv-table:: **Example Ranking Table**
```
dcdenu4, line 58: Just a nit-pick here, but can we get a little space between the table and 
>The model calculates the exposure...

source/wave_energy.rst
```diff
@@ -59,32 +59,25 @@ We prepared globally and regionally available input data layers for the potentia
...
 + AquaBuOY (Dunnett and Wallace 2009)
 + WaveDragon (Dunnett and Wallace 2009)
-
-.. Table 3.2 shows an example of wave energy absorption performances in each seastate bin for Pelamis.
  
 .. figure:: ./wave_energy/table_energyabsorption.png
    :align: center
    :figwidth: 500px
 
-   Wave energy absorption performance (kW) in each seastate bin for Pelamis.
```
dcdenu4, line 87: Did we want to keep this in?

dcdenu4 (changes requested):
Hey @emlys ,

Thanks for making another pass at this and fixing some of the table numbering issues. I just a few small comments going through this again. But it's looking pretty much good to go.

dcdenu4 (approved):
Alright @emlys let's make this canonical!

PR #24, opened 2021-03-23
Stormwater

emlys:
User's guide for the new stormwater model.


emlys:
This is on hold while @phargogh is away. Ideally we'll merge this, then update the UG hash in natcap/invest#523, then merge that.

emlys:
Sorry if you get a notification @chrisnootenboom! Accidentally clicked on you :)

source/stormwater.rst
```diff
@@ -0,0 +1,329 @@
...
+Users may provide a polygon vector file outlining areas over which to aggregate data (typically watersheds or sewersheds). The model will aggregate the output rasters to compute:
+
+- Average stormwater retention ratio (average of :math:`RE` values)
+- Total retention volume, :math:`m^3` (sum of :math:`VRE` values)
+- Total retained pollutant load for each pollutant, :math:`kg/yr` (sum of :math:`Avoided.load` values)
+- Total pollutant load for each pollutant, :math:`kg/yr` (sum of :math:`Load` values)
+- Total potential recharge volume, :math:`m^3` (sum of :math:`VI`, if infiltration data provided)
+- Total Replacement Cost, currency units (sum of retention costs, if value data provided)
+
+
```
emlys, line 108: Would be nice to get a "Limitations and Simplifications" section in here like we have for other models. Specifically this could address the seemingly high valuation outputs, which I think Ben said could be because most reporting uses the export portion.

source/stormwater.rst
```diff
@@ -0,0 +1,329 @@
+.. _stormwater:
+
+***********************************************
+Stormwater Retention Model
```
emlys, line 4: Given that Milan brought it up, let's revisit the model naming question. "Urban Rain Model" or "Urban Water Model" sound good to me, to highlight how it's different from the Urban Flood model. Suggestions welcome! I'm inclined to leave out "Retention" since that is only one of several outputs.

source/stormwater.rst
```diff
@@ -0,0 +1,329 @@
...
+***********************************************
+
+
+Overview
+========
+Planning agencies are increasingly considering urban water management in their strategies to address issues such as climate change, flood risk, or population growth, while protecting the environment. The InVEST stormwater runoff retention model can support this effort by providing information on two ecosystem services related to stormwater management: runoff retention, and groundwater recharge (Flood risk reduction is assessed in a separate InVEST model, :ref:`Urban Flood Risk Mitigation <ufrm>`). Runoff retention corresponds to the retention of stormwater by pervious land uses, which is beneficial given the detrimental effects of polluted stormwater discharge into rivers or the ocean. Groundwater recharge is a related service, corresponding to the percolation of stormwater past the root zone, potentially recharging groundwater for human and non-human purposes.
```
emlys, line 10: Feedback from Milan:

> Considering the potential similarity between the stormwater model and the urban flood model, I feel like it would help to mention in the beginning of the model description maybe the main points in how they differ, to make the decision for which model to use easier for the potential end user.
This could for example include explaining why ‘retention costs’ can be calculated with the stormwater model but not with the flood risk model? Or why one model uses the Curve Number method, but the other uses Runoff Coefficients (which I can imagine users are likely to mix up if they use both models). Both questions I also still ask myself.

Depending how much there is to say, this could also go in a "Comparison to Urban Flood Model" section.

source/stormwater.rst
```diff
@@ -0,0 +1,329 @@
...
+- **search_kernel.tif**: A binary raster representing the search kernel that is convolved with the retention_ratio raster to calculate the averaged retention ratio within the retention radius of each pixel
+
+- **ratio_average.tif**: A raster where each pixel's value is the average of its neighborhood of pixels in the retention_ratio map, calculated by convolving the search kernel with the retention ratio raster
+
+
+.. _Input Guidance:
+
+Input Guidance
+==============
+
```
emlys, line 219: Feedback from Milan:
> The retention radius parameter is completely user definable, but no ‘default’ value is given. I have some trouble finding an adequate value in literature for this. Could you maybe provide a default value for this setting, or a literature reference to learn more about a suitable value?

We should ideally provide some guidance on every input.

source/stormwater.rst
emlys: The model name is not difficult to change in the code, if we end up changing it.

source/stormwater.rst
```diff
@@ -0,0 +1,329 @@
...
+- **Soil hydrologic group** (required): Raster map of soil hydrologic groups. Each pixel's value is 1, 2, 3, or 4, corresponding to hydrologic soil groups A, B, C, and D respectively.
+
+- **Precipitation** (required). Raster map of annual precipitation in :math:`mm/yr`
+
+- **Biophysical table** (required). A .csv (Comma Separated Value) table containing model information corresponding to each of the land use classes in the LULC raster. *All LULC classes in the LULC raster MUST have corresponding values in this table.* Each row is a land use/land cover class and columns must be named and defined as follows:
+
+    - **lucode**: Unique integer for each LULC class (e.g., 1 for forest, 3 for grassland, etc.) *Every value in the LULC map MUST have a corresponding lucode value in this column.*
+
+    - **is_impervious** (only required if **Adjust retention ratios** is checked: Binary value indicating whether each LULC class is directly-connected impervious (enter 1 for yes, 0 for no). This is used to adjust the retention coefficients for nearby pixels.
```
emlys, line 126: Feedback from Milan:
> For the ‘is impervious’ column, it might need to be a bit more clear when a LULC is considered to be impervious enough to be labelled as 1 for the model. Is every class that is not considered 100% pervious (e.g. grassland) considered pervious? (e.g. very low density urban fabric with 20% perviousness?) Or is it meant to ‘assume’ every urbanized area as ‘impervious’ due to their likely connection to the sewage system? (I already adjusted the RC and IR values for each LULC depending on their avg. imperviousness so that is not an issue)

Ben said via email:
> This is where the documentation could use some more detail, as the consideration of connectedness is somewhat subjective, but also tied to the land cover layer being used. We developed it with the US National Land Cover Database (NLCD) in mind, which has some broad "developed" categories (Open, Low, Medium, and High) that are essentially bins of imperviousness. We further made the assumption that the "High" and "Medium" categories would likely be fully connected to storm sewer systems (due to having > 50% imperviousness per the NLCD specification), while the Low and Open categories were lower imperviousness (< 50%) and might contain some cover types like parks that are not connected and/or provide more retention. We left the determination of "connectedness" up to the user, since land cover layers might have more (or less) detail, depending on where the model is being used.

This is great info to include.

emlys:
@phargogh, I think we're still waiting on the science team.

emlys:
closing and moving to a new branch - diff got messed up somehow

PR #25, opened 2021-04-30
Updates to Urban Cooling UG chapter

jagoldstein:
substantial edits made to "urban_cooling_model.rst" and "ucm_appendix_table.csv" to improve grammar, readability, and language consistency


jagoldstein:
I reviewed the HTML and CSV files. They both look good and this PR is ready to be merged.

source/urban_cooling_model.rst
```diff
@@ -159,104 +158,103 @@ For each pixel, the model computes the estimated loss in productivity, in %, for
...
 
-* Reference Air Temperature (:math:`T_{ref}`): Rural reference temperature (where the urban heat island effect is not observed) for the period of interest. This could be nighttime or daytime temperature, for a specific date or an average over several days. The results will be given for the same period of interest).
+* Baseline air temperature (:math:`T_{ref}`): Rural reference air temperature (where the urban heat island effect is not observed) for the period of interest. This could be nighttime or daytime temperature, for a specific date or an average over several days. The results will be given for the same period of interest.
 
-* Magnitude of the UHI Effect (:math:`UHI_{max}`) : Magnitude of the urban heat island effect, in degrees Celcius, i.e. the difference between the rural reference temperature and the maximum temperature observed in the city.
+* Magnitude of the UHI effect (:math:`UHI_{max}`): Magnitude of the UHI effect (in degC), i.e. the difference between the rural reference (baseline air) temperature and the maximum temperature observed in the city.
```
emlys, line 210: I kind of like using the unicode degree symbol (°) where possible. It's probably fine but I wonder if "degC", especially run together as one word, might be confusing to users who aren't native English speakers. Spelling out fully and/or using SI conventions may be more universally understandable.

source/urban_cooling_model.rst
```diff
@@ -159,104 +158,103 @@ For each pixel, the model computes the estimated loss in productivity, in %, for
...
 
-* Air Temperature Maximum Blending Distance: Search radius (in meters) used in the moving average to account for air mixing (default value: 2000m)
+* Air Temperature Maximum Blending Distance: Search radius (in meters) used in the moving average to account for air mixing (default value: 600m).
```
emlys, line 212: The UI does set a default of 2000 for this input [here](https://github.com/natcap/invest/blob/01a3ddeac94d60afd124542b15503fdf916abb0f/src/natcap/invest/ui/urban_cooling_model.py#L75) .However, I'm not sure about the concept of a "default" value to begin with. Here are some problems I see with it: 
- When you save the inputs from the last model run, any defaults are overridden. As far as I know there isn't a way to "reset defaults" in the current UI.
- This is because a "default" is handled entirely on the UI side: it's simply a string that's put into the input field when it first loads.
- There is no representation of a "default value" in the `ARGS_SPEC`, and I'm not convinced that there should be.
- Because it's not represented in the `ARGS_SPEC`, it will not carry over to the workbench UI.
- All a "default" really means is a commonly-accepted value that may be suitable for a lot of scenarios. Officially encoding this into the model spec would introduce some questions with unclear answers: How common does a value have to be, to get the designation of "default"? How do we decide which inputs get defaults? Are we accidentally misleading users or promoting certain papers/methods by giving extra weight to a "default" value?

All that is to say that I don't really like the term "default", and I might say something like "A common value is 2000 m (citation)" instead. @phargogh do you have any thoughts about defaults?

source/urban_cooling_model.rst
```diff
@@ -159,104 +158,103 @@ For each pixel, the model computes the estimated loss in productivity, in %, for
...
 
-* Cooling capacity calculation method: Either "Weighted Factors" or "Building Intensity".  The method selected here determines the predictor used for air temperature.  If "Weighted Factors" is selected, the Cooling Capacity calculations will use the weighted factors for shade, albedo and ETI as a predictor for daytime temperatures.  Alternatively, if "Building Intensity" is selected, building intensity will be used as a predictor for nighttime temperature instead of shade, albedo and ETI.
+* Cooling Capacity Calculation Method: Either "Weighted Factors" or "Building Intensity". The method selected here determines the predictor used for air temperature. If "Weighted Factors" is selected, the CC calculations will use the weighted factors for shade, albedo, and ETI as a predictor for daytime temperatures. Alternatively, if "Building Intensity" is selected, building intensity will be used as a predictor for nighttime temperature instead of shade, albedo, and ETI.
```
emlys, line 214: Hooray for the oxford comma! I want this in our style guide 🤓

emlys (changes requested):
Thanks for this @jagoldstein! Definitely a good improvement and all our work will seem more reputable when we follow good writing style guidelines. 

To avoid us duplicating efforts, I might suggest holding off on doing this for other models for now. Urban Cooling is pretty unique. But some other model user guides have entire sections that are repeated word-for-word in other models (e.g. https://storage.googleapis.com/releases.naturalcapitalproject.org/invest-userguide/latest/ndr.html#appendix-data-sources) or reuse common phrases/explanations ("Make sure that there is ample disk space and that write permissions are correct.", ".csv (Comma Separated Values)"). All of these I'd like to put in one centralized file where we go over common inputs, outputs, file type basics, etc., and that would happen before the translations.
There is also a lot of overlapping text in the `"about"` property of each model `ARGS_SPEC` that will need to be merged with user's guide descriptions (again to reduce translation costs, and so we can automatically generate those portions of the user's guide). So it might make sense to do this kind of editing for other models after those changes happen.

source/urban_cooling_model.rst
```diff
@@ -7,13 +7,13 @@ Urban Cooling Model
...
-Natural infrastructure therefore plays a role in reducing the urban heat island in cities. Using the rapidly-growing literature on urban heat modeling (Deilami et al., 2018), the InVEST urban cooling model estimates the cooling effect of vegetation based on commonly available data on climate, land use/land cover (LULC), and (optionally) A/C use.
+UHIs affect many cities around the world, with major consequences for human health and wellbeing: high mortality or morbidity during heat waves, high A/C consumption, and reduced comfort or work productivity. The UHI effect, i.e. the difference between rural and urban temperatures, is caused by a change in the energy balance in cities due to two main factors: the thermal properties of materials used in urban areas (e.g. concrete, asphalt), which store more heat, and the reduction of the cooling effect of vegetation (through shade and evapotranspiration).
```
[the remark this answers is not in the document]
davemfish: I think the "UHI" acronym needs to be spelled out on it's first use.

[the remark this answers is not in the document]
davemfish: My naive thinking on energy balance is just "heat going in vs heat going out". But sentence could also read:

"...is caused by 1) a change in the thermal properties of materials used in urban areas (e.g. concrete, asphalt), which store more heat, and 2) the reduction of the cooling effect of vegetation (through shade and evapotranspiration).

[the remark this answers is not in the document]
jagoldstein: It's just a general term used in thermodynamics, but I agree that it may be overly technical for this sentence
```suggestion
UHIs affect many cities around the world, with major consequences for human health and wellbeing: high mortality or morbidity during heat waves, high A/C consumption, and reduced comfort or work productivity. The UHI effect, i.e. the difference between rural and urban temperatures, is a result of the unique characteristics of cities due to two main factors: the thermal properties of materials used in urban areas (e.g. concrete, asphalt), which store more heat, and the reduction of the cooling effect (through shade and evapotranspiration) of vegetation.

```

source/urban_cooling_model.rst
```diff
@@ -24,9 +24,9 @@ How it works
...
-The shade factor ('shade') represents the proportion of tree canopy (for trees >2m) associated with each land use/land cover (LULC) category. Its value is comprised between 0 and 1.
-The evapotranspiration index represents a normalized value of potential evapotranspiration, i.e. the evapotranspiration from vegetation (or evaporation from soil, for unvegetated areas). It is calculated for each pixel by multiplying the reference evapotranspiration (:math:`ET0`, provided by the user) and the crop coefficient (:math:`Kc` , associated with the pixel's LULC type), and dividing by the maximum value of the :math:`ET0` raster in the area of interest, :math:`ETmax`.:
+The model first computes the cooling capacity (CC) index for each pixel based on local shade, evapotranspiration, and albedo. This approach is based on the indices proposed by Zardo et al. 2017 and Kunapo et al. 2018, to which we add albedo, an important factor for heat reduction.
+The shade factor ('shade') represents the proportion of tree canopy (≥2m in height) associated with each land use/land cover (LULC) category. Its value is comprised between 0 and 1.
```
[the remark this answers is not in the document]
jagoldstein: Yes, this is an intentional change to more precise language. See Section 2.2 from [Zardo et al., 2017](https://doi.org/10.1016/j.ecoser.2017.06.016), "_trees with canopy equal or higher than two meters_". Nice catch though, thanks for asking!

source/urban_cooling_model.rst
```diff
@@ -53,17 +53,17 @@ Optionally, the model can consider another factor, intensity (:math:`building.in
 Urban heat mitigation index (effect of large green spaces)
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-To account for the cooling effect of large green spaces (>2 ha) on surrounding areas (see discussion in Zardo et al., 2017 and McDonald et al. 2016), the model calculates the urban heat mitigation (HM) index: HM is equal to CC if the pixel is unaffected by a large green spaces, and otherwise set to a distance-weighted average of the CC values from the large green space and the pixel of interest.
+To account for the cooling effect of large green spaces (>2ha) on surrounding areas (see discussion in Zardo et al., 2017 and McDonald et al. 2016), the model calculates the urban HM index: HM is equal to CC if the pixel is unaffected by any large green spaces, but otherwise set to a distance-weighted average of the CC values from the large green spaces and the pixel of interest.
```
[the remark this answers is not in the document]
jagoldstein: Thanks for bringing this up and for being so detail focused. I don't have a preference for space or not, but could easily be convinced that using one is best. Looking through the UG, I see it both ways, often within the same model chapter, so YES, as long as it's consistent throughout, I'd be happy. +1 for this being 
> part of a different, style-focused PR that touches all the models

source/urban_cooling_model.rst
phargogh: > All that is to say that I don't really like the term "default", and I might say something like "A common value is 2000 m (citation)" instead.

Yes, I agree that "default" implies that the value is a fallback, which isn't quite correct.  I think "recommended" value is maybe closer to what we really mean ... it's more like a "this value should be sufficient for most use cases, but we want you to be able to tweak it if you know what you're doing".

RE: citations, these would be great to include with a recommended value where available!  But this should be able to handle 0+ citations.  A concrete use case for this is that RouteDEM allows the user to provide a band index (recommended value: 1).  DelineateIt has a checkbox marked as True (whether to skip invalid geometry).  No citations needed there.

> There is no representation of a "default value" in the ARGS_SPEC, and I'm not convinced that there should be.

This is true!  We've been shoehorning a default value in to the UI files or just using the parameter set files to represent what might be better represented as default/recommended values.

I have 2 reasons why I think we should seriously consider adding a recommended value to the ARGS_SPEC for non-file inputs:
1. Our science staff sometimes ask for them!  At times the request is better represented as part of a parameter stack, but there do seem to be some 'expert opinion' cases where a natcap-scientist recommended value might be good enough for most.
2. For cases where the model has a recommended or default value anyways (urban cooling example within `execute`), it'd be better to have a single source of truth for that value by referring back to the ARGS_SPEC.  This could also enable us to have a more consistent user experience with recommended/default values across models, too.

If we were to have ARGS_SPEC be the source of truth for recommended values, it would also solve the other concerns you mentioned about values not being available to the workbench and would allow for an individual value to be reset to its original recommended/default value.

source/urban_cooling_model.rst
jagoldstein: I agree. I somewhat lazily choose 'degC' here this time to make it short and consistent because there were a variety of spellings including an incorrect one ("Celcius"; plus a couple "centigrade"s). I think the unicode symbol (°) would be better than 'degC' for the reasons @emlys describes. SI conventions and/or fully spelling out units would be better too (and more inclusive). Again, I could be convinced of almost any way here, as long as we establish a standard and stick to it.

source/urban_cooling_model.rst
jagoldstein: Great points @emlys , especially about maybe just using a different wording, like "common value" or "recommended starting value" rather than "default".
Specific to this case, please see [this Forums post](https://community.naturalcapitalproject.org/t/urban-cooling-queries-about-parameterization/915) from April 2020 where @royremme wrote,
> "I suggest starting with the 500m value in the appendix that is based on the first studies we have used the model in. The 2000m is an artifact from an older version of the user guide, we need to update that".
That doesn't quite explain my "600m", but I believe this value came from what @newtpatrol and @chrisnootenboom have used most recently for the Twin Cities and what we recommended in the data packet we used for Gretchen's class.

source/urban_cooling_model.rst
jagoldstein: Yes, Oxford comma love! I cringe at lists without it :fearful:.

BUT for the translations, be aware that, in my experience, the Oxford comma is (universally?) absent from the Spanish language.

emlys: Oh, good to know!

phargogh: That is interesting to know about the Spanish language!

When all of the text here is translated, it'll be up to the translator to translate each chunk (in this case, it'll most likely be a paragraph) in a way that makes sense for the target language, and the text will be swapped in paragraph-by-paragraph.  So for us, we should for sure make decisions about English grammar or style that make sense for English and then do the same for Spanish when we translate it into Spanish.  If the two paragraphs end up having a _slightly_ different arrangement of commas, that'll be ok!

source/urban_cooling_model.rst
phargogh: >>    "I suggest starting with the 500m value in the appendix that is based on the first studies we have used the model in. The 2000m is an artifact from an older version of the user guide, we need to update that".
>
>That doesn't quite explain my "600m", but I believe this value came from what @newtpatrol and @chrisnootenboom have used most recently for the Twin Cities and what we recommended in the data packet we used for Gretchen's class.

@jagoldstein is it your impression that this value will be specific to the site, or do you think this will be a pretty common value across study areas?  If it's for the sake of demonstration, I'd think this might be better-suited to a parameter set (demonstrating this numeric value in the context of other datasets that made sense for that analysis) than a standalone recommended value.  What do you think about that?

source/urban_cooling_model.rst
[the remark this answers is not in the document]
jagoldstein: I agree that acronyms should be spelled out on first use and then not again. The first occurrence of "urban heat island" is in the 'Summay' section that precedes the 'Introduction', so that is where I first defined "UHI".

source/urban_cooling_model.rst
jagoldstein: @phargogh great points and questions. It is my impression that there can be relatively high uncertainty in the max air blending distance, as described by @chrisnootenboom in [this Forums post](https://community.naturalcapitalproject.org/t/air-temperature-maximum-blending-distance-uncertainty/1795/3). Chris says that 600m comes from "Shatz and Kucharik papers" (not cited in UG). The model was tested in the Twin Cities and Paris, France before release, but _not_ in significantly different locations, like the tropics. I think your idea of a parameter set sounds like it could make more sense than "recommended" or "default" values. I should defer to Chris or Eric though as I am far from an expert on this.

source/urban_cooling_model.rst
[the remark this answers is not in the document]
jagoldstein: I'm convinced @emlys . I'm going to recommend [this standard](https://developers.google.com/style/units-of-measure) for our [style guide.](https://docs.google.com/document/d/1BHwHDu_I-x0s_2GsbUb4rfVmXMkl7kl97sx2suBTLh8/edit?usp=sharing)

source/urban_cooling_model.rst
chrisnootenboom: @jagoldstein @phargogh I just caught up on this thread and can confirm: the 500m value was derived from regression modeling in the Schatz and Kucharik paper (https://doi.org/10.1175/JAMC-D-14-0107.1) while the 600m value came from similar regression modeling in our golf paper (https://doi.org/10.1016/j.landurbplan.2020.104022). Users should be aware of this limitation when applying the model to other climate zones, but there just aren't many data sources on this value out there. 
It is unclear whether the parameter will be hyper-specific to an individual city--air flow dynamics should be consistent globally, but our current calibration studies in the Twin Cities and Paris result in a range of values from as low as 260m to as high as 1600m. So the jury is definitely out on a single 'default' value even in cities in temperate climates!
That being said, I would reference the two papers I linked above in the users guide if possible, since 500 or 600m is a reasonable first pass for the model.

source/urban_cooling_model.rst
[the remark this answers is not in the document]
emlys: Neat! That google style guide looks perfect.

source/urban_cooling_model.rst
jagoldstein: Thanks for this info @chrisnootenboom ! I adjusted the language accordingly and added these two sources to the References section.

source/urban_cooling_model.rst
```diff
@@ -159,104 +159,103 @@ For each pixel, the model computes the estimated loss in productivity, in %, for
...
 
-* Reference evapotranspiration: a raster representing reference evapotranspiration (units of millimeters) for the period of interest (could be a specific date or monthly values can be used as a proxy)
+* Reference Evapotranspiration: A raster representing reference evapotranspiration (units of millimeters) for the period of interest (could be a specific date or monthly values can be used as a proxy).
 
-* Areas of interest: polygon vector delineating areas of interest (city boundaries or neighborhoods boundaries). Results will be aggregated within each shape contained in this vector
+* Area of interest: Polygon vector delineating areas of interest (city boundaries or neighborhoods boundaries). Results will be aggregated within each shape contained in this vector.
 
-* Green Area Maximum Cooling Distance (:math:`d_{cool}`) : Distance (in meters) over which large urban parks (> 2 ha) will have a cooling effect
+* Green Area Maximum Cooling Distance (:math:`d_{cool}`): Distance (in meters) over which large urban parks (>2ha) will have a cooling effect (recommended value: 450m).
```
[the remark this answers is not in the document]
jagoldstein: I agree about needing those spaces before the unit. >2 ha is correct. The change to 'greater than or equal to' was for the 2 m tree canopy height, not the 2 ha green space area.

davemfish:
@jagoldstein I haven't looked at this lately but it looks like Emily has given a pretty thorough review. Let me know if there is anything specific I can weigh-in on, otherwise I think Emily's review is probably sufficient and maybe I can be off-the-hook for reviewing this one?

jagoldstein:
You are off the hook @davemfish ! @emlys is all over it. I just wasn't sure of the best way to proceed after committing further changes following the first round of reviews. One thing I learned is NOT to resolve comments (because it seems I can not retrieve them after that).

source/urban_cooling_model.rst
```diff
@@ -122,26 +122,26 @@ Where:
...
+**Work Productivity:** the model converts air temperature into Wet Bulb Globe Temperature (WBGT) to calculate the impacts of heat on work productivity. WBGT takes into account humidity, and can be estimated from standard meteorological data in the following way (American College of Sports Medicine, 1984, Appendix I):
 
 .. math:: WBGT_i = 0.567 \cdot T_{air,i} + 0.393 \cdot e_i + 3.94
     :label: [7]
 
 Where:
 
-    * :math:`T_{air}` = temperature provided by the model (Dry bulb temperature (:math:`T_C` ))
-    * :math:`e_i` = Water vapour pressure (hPa) [humidity]
+    * :math:`T_{air}` = temperature provided by the model (dry bulb temperature (degC))
```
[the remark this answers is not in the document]
jagoldstein: good catch... I'm going with `° C`, per your earlier suggestion and our current style guide

source/urban_cooling_model.rst
```diff
@@ -159,104 +159,103 @@ For each pixel, the model computes the estimated loss in productivity, in %, for
...
-    * Building_intensity: A floating-point value between 0 and 1.  This is calculated by dividing the floor area by the land area, normalized between 0 and 1.  Required if using the weighted factor approach to Cooling Coefficient calculations.
+    * lucode: Required. LULC class code. Codes must match the 'value' column in the LULC raster and must be unique integer or floating point values, in consecutive order.
+    * Shade: A value between 0 and 1, representing the proportion of tree cover (0 for no tree; 1 for full tree cover with canopy ≥2m in height). Required if using the weighted factor approach to CC calculations.
+    * Kc: Required. Crop coefficient, a value between 0 and 1 (see Allen et al. 1998).
+    * Albedo: A value between 0 and 1, representing the proportion of solar radiation directly reflected by the LULC class. Required if using the weighted factor approach to CC calculations.
+    * Green_area: Required. A value of either 0 or 1, 1 meaning that the LULC class qualifies as a green area (green areas >2ha have an additional cooling effect), and 0 meaning that the class is not counted as a green area.
```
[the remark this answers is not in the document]
jagoldstein: `>2 ha`

jagoldstein:
@emlys I believe I made all of the needed changes that you identified. Thanks again for being so thorough! I hope this PR is ready for merging now 🤞

emlys (approved):
Awesome thank you @jagoldstein for putting up with my picky suggestions! 😁

PR #29, opened 2021-06-03
Build independently and within InVEST build

### LINKED ISSUE #28: Make sample data work within InVEST builds
Originally pointed out by @richpsharp. When the UG commit hash got updated to a more recent version that includes the changes from https://github.com/natcap/invest.users-guide/pull/21, it broke the InVEST binary builds. That's because the UG build running within the InVEST build expected to have its own copy of the sample data, but it doesn't.

Need to handle both cases:
- Independent UG build (like in github actions on natcap/invest.users-guide). It should check out its own copy of the sample data.
- UG build within the InVEST build (in github actions on natcap/invest). It should ideally use the copy of the sample data already existing in the InVEST directory.

There will be changes to both the invest makefile and the user's guide makefile. The corresponding ticket for invest is here: natcap/invest#559

emlys:
Fixes #28

Some constraints:
- paths in .rst files can't be absolute because the path above the invest.users-guide root directory can change. So, they have to be relative to the .rst file location.
- paths in .rst files are static (unless we made a custom sphinx directive). They need to point to the sample data files whether we're building the UG by itself or within an InVEST build.
- The UG build process shouldn't assume or change anything about what's outside of the invest.users-guide directory.

The `make html` command now assumes that a clone of `invest-sample-data` already exists in the top level of `invest.users-guide`.
If we're building the UG by itself, it checks out its own copy of `invest-sample-data` before calling `make html`.
If we're building the UG within an InVEST build, the InVEST build is responsible for making sure that the `invest-sample-data` directory exists where the UG expects it. This is implemented in natcap/invest#571 with a symlink.

This PR should be merged before natcap/invest#571 so that we can update the invest PR to point to natcap/invest.users-guide before merging.


davemfish (changes requested):
This looks great, the build looks great. One minor thing is that a couple of the make targets are missing from the "help" target. Would you like to add those @emlys ?

emlys:
Thanks for catching that @davemfish. I added them to the help text and also updated the README.

davemfish (approved):
Looks great!

Issue #31, opened 2021-06-10
Translations: Consolidate text betwen the user's guide and args specs


emlys:
In general, descriptions of model inputs will be shifted into the args specs and replaced with the custom sphinx role developed in #30.

emlys:
This was done in #42, #45, and #47.

PR #36, opened 2021-07-16
Consolidate text between UG and args spec (part 1 of 3)

emlys:
#31, models A-D


requirements.txt
```diff
@@ -1,6 +1,8 @@
 # These are the packages that need to be installed in order to build HTML
 # documentation of the User's Guide.
 
+git-lfs
+natcap.invest
```
phargogh, line 5: Hey, I know this PR is still in draft, but I noticed this line and thought this might be useful: since the user's guide is intended to be a continuously-updated resource and we're now building parts of the UG off of the state of InVEST, what would you think about installing `natcap.invest` from the latest git source ([pip docs](https://pip.pypa.io/en/stable/cli/pip_install/#git))?

emlys:
I am closing this PR because the GHA workflow has not run in weeks for no apparent reason. I'll open a new PR and see if that triggers it to run.

Issue #16, opened 2021-01-27
Equation numbers are off


dcdenu4:
I noticed in the NDR and other models that equation number are in the 20s, or 40s... This doesn't seem right.

dcdenu4:
We talked about equation numbers in the UG on the software team call recently. We talked about
- Should each chapter restart equation / figure numbers? Or, as is currently done, do we like the running equation numbers across chapters?
- Are there better ways to reference equations outside the user guide for the ineveitability of those equation numbers changing. See #194 

We should call in @newtpatrol , @jagoldstein , @lmandle to this conversation too!

lmandle:
+1 to Jesse's preference for restarting numbers within each chapter. In my experience, most people who read the user's guide are looking the chapters individually for the model(s) they are using rather than reading sequentially.

PR #37, opened 2021-07-27
Update SDR and NDR natcap/invest#555

emlys:
Goes with natcap/invest#555.

- Added description of the new pour points feature
- Added a bunch of intermediate outputs that were missing from the Interpreting Results section
- Added equations for the thresholded and bare soil outputs to distinguish them from the others, and referenced those equations in the Interpreting Results section
- Added years to the units for rainfall erosivity. It was unclear how yr^-1 got into the USLE from the R, K, LS, C, P units, but sources indicate erosivity is per-year (e.g. https://esdac.jrc.ec.europa.eu/themes/rainfall-erosivity-europe)


emlys:
Hey @jagoldstein and @newtpatrol, would one or both of you have time to look over these changes? The Pour Points section was specifically related to an invest update that'll be released in 3.9.1, but also made other changes with the goal of clarifying some stuff that confused me. 

I feel like the subject of "sediment retention" vs "sediment retention index" came up recently but I can't remember what was concluded. I added a separate section for Sediment Retention that just has the equation right now. I'm not sure what to add as far as interpretation/how it's different from Sediment Retention Index.

newtpatrol (commented):
Good catch! Yes, the units for erosivity should be per year.

emlys:
@newtpatrol Sadly I don't think there's a way to delete comments :( but you can edit it to say "deleted" or something

source/sdr.rst
```diff
@@ -46,7 +46,7 @@ The amount of annual soil loss on pixel :math:`i`, :math:`usle_i` (units: :math:
 
 where
 
- * :math:`R_i` is rainfall erosivity (units: :math:`MJ\cdot mm (ha\cdot hr)^{-1})`,
+ * :math:`R_i` is rainfall erosivity (units: :math:`MJ\cdot mm (ha\cdot hr\cdot yr)^{-1})`,
 
  * :math:`K_i` is soil erodibility (units: :math:`ton\cdot ha\cdot hr (MJ\cdot ha\cdot mm)^{-1}`),
```
jagoldstein, line 51: Are the units of erodibility correct _(hours_ versus _years_)? Why is _ha_ in both the denominator and the numerator when they would cancel out?

source/sdr.rst
```diff
@@ -104,21 +117,73 @@ Sediment Delivery Ratio
...
+Pour Points
++++++++++++
+
+:math:`d_i` is the distance that water travels downslope from the pixel before reaching a stream. Because SDR is calculated in terms of :math:`d_i`, it is only defined for pixels that drain into a stream on the map.
+
+In order to ensure that every pixel drains to a stream, and thus has defined :math:`d_i` and SDR values, the model adds pour points to the streams map if needed. A pour point is a point where water flows off the defined area of the flow direction raster, either off the edge or into a nodata area. If these points are not already marked as streams in the streams layer generated from flow accumulation (**flow_accum_streams.tif**) or the optional drainage layer input, the model will mark them as streams in **streams.tif**. These extra stream pixels may be a sign that something is wrong with your inputs:
+
+- If the pour point is outside your watersheds of interest, you may ignore it.
+
+- If the pour point is actually a stream in the real world, try increasing your TFA and/or using the optional drainage layer so that the model recognizes this stream.
```
jagoldstein, line 164: I'm confused here. I thought this was a list of how to deal with the model defining more pixels as streams than it should, so why would the user make a change "_so that the model recognizes this [more] stream_"? And increasing the TFA would lead the model to recognize fewer stream pixels anyway, not more.

source/sdr.rst
```diff
@@ -104,21 +117,73 @@ Sediment Delivery Ratio
...
+
+- If the pour point is not a stream in the real world, try using larger input data to be sure it completely covers the watershed including its streams.
```
jagoldstein, line 166: What are "_larger input data_"? Do you mean "_input data with a greater extent_"?

jagoldstein (changes requested):
These updates generally look good to me and include some important corrections. I added a few inline comments with questions I have, but it is difficult for me to visualize the math changes in GitHub and know if they are correct or not without diving deeper into the source code.
Why were the [_Suffix]s removed from all the filenames? The outputs include them and are not actually named "_stream.tif_", for example.

source/sdr.rst
newtpatrol: @jagoldstein, the units for erosivity are weird. As I commented previously: The correct units are listed in the Data Needs section: (MJ-mm)/(hour-ha-year). Also see this document, which walks through the gnarly units involved in the USLE: https://www.engr.colostate.edu/~pierre/ce_old/Projects/linkfiles/USLE%20Unit%20conversions.pdf.

jagoldstein: Thanks, yeah the units get weird for both erodibility and erosivity, and I see different variations across different sources. I was specifically checking on _soil erodibility_ units here, which Emily did not change the definition of yet, but I just wanted to confirm that what's written for that is accurate.

emlys:
@jagoldstein, to see the built HTML docs, you can download them from Github: In the "Checks" tab, in the top right corner there is an "Artifacts" button. Click on that and you can download a zip file of the complete built UG as of the latest change. I wish we had a way to view it directly in the browser, hmm...

source/sdr.rst
emlys: It's strange that the _ha_s don't cancel, but that paper @newtpatrol linked says:
> Although hectares  in  the numerator cancels hectares in the denominator,  both  are left  to show that  K  is  soil loss  per  unit area per unit of  E I .   
It also shows _hours_ for erodibility (see the table on page 3)

source/sdr.rst
emlys: You're right it should say _decreasing_ TFA, not increasing, to make more streams. Hooray for reviews! 🎉 

My understanding is that, the user should provide data such that every pixel in their DEM drains into a stream.

When that doesn't happen, as of 3.9.1 the model will handle it by adding stream pixels on the pour points. But really, every pour point should _already_ be marked as a stream (either in the the stream layer that the model generates, and/or the optional streams input layer). 

Therefore, added stream pixels are a sign that you should mark more streams in that area, either by lowering the TFA or providing the optional streams input layer. Does that sound right?

jagoldstein: Ok, that makes more sense to me. @newtpatrol do you agree?

source/sdr.rst
jagoldstein:
```suggestion
- If the pour point is not a stream in the real world, try inputting data with a greater spatial extent to be sure it completely covers the watershed including its streams.

```

emlys: That does read better, thanks!

emlys:
> In general, I don't think we provide enough explanatory text with the outputs, really we should probably provide example graphics too, since we get so many questions from people "does this look ok?"

Yes!! I eventually hope to have information about intermediate and final outputs encoded like we do for inputs in the `ARGS_SPEC`, and generate this part of the docs as well (natcap/invest#596)

> Regarding sediment retention and sediment retention index, YES that should be clarified. I still do not understand the difference, and maintain that if we're providing something that is an "index" it should be on a scale of 0-1. 

I agree that an index should be 0-1, and need to get definite clarification on what these outputs mean (and if that's not possible to get, consider deprecating one/both).  Providing the equation with no explanation isn't great, but maybe the best we can do for now.

> Given that I don't currently understand the difference, I'd prefer to have a single sediment retention index provided (not sed_retention, but an index), on a scale of 0-1, which makes it clear that it is not to be interpreted quantitatively, which we've had to deal with on the forum many times.

I made an issue for this including your comments: https://github.com/natcap/invest/issues/601 

> Has it always been the case that we threshold cover-management factors? I'm a little confused by that, since it would override the user's hopefully carefully-researched input.

I believe it's always been the case, the thresholding step dates back at least 3 years ago in our commit history. I just added the equations to better line up with what the model does. I agree seems strange to threshold the C values. Why not simply require all C values in the biophysical table to be greater than 0? 

> I'd change to "it could be a sign that something is wrong with your DEM", not inputs in general.
> Expand TFA acronym to threshold flow accumulation.

👍 

> I'm not entirely sure about this pour point addition, so this may be off, but it makes me think about why this would happen. It could happen when people have a buffer of input data that extends past the limits of the input watersheds. In this case, they may ignore it. However, if it's the case that they actually mismatched their watershed and the actual drainage produced from the DEM (which happens when people don't create their own watershed from the same DEM, but use one that was created elsewhere), then that is a problem, and they should re-evaluate their watershed.

I was thinking about the "buffer" case here. Thanks for pointing out the mismatched watershed situation!

> In general, I'm not sure about this handling of pour points. In some ways it seems ok. For example, in the past I've intentionally created NoData areas in a DEM as a simple way to force water to drain to lakes. In this case, it might have the desired effect. But in the case where pixels are flowing off the map, that's not really export, that's just unknown and probably incorrect, and seems like it could lead to edge weirdness in the results. This is also the case where it's being caused by bad or gnarly DEM data.

I am also not sure about handling it this way. It is definitely an improvement, because before, areas that did not drain to a stream would fill with meaningless values. The pour points thing was intended to solve that problem without creating NoData areas. Here's how Rich explained the rationale to me:

> I've encountered this problem in my work for watersheds that just aren't big enough to define a stream, but i want one anyway. I've gotten around this by not only defining a channel layer but also detecting single pixel outlets and setting those to be "channels" as well.
my feeling is that SDR should have a defined value in these areas otherwise you'll get users complaining that there are large swaths of nodata values in their results and then you have to explain that their stream threshold is too high or whatnot.
but i don't think that by itself the mfd distance to channel function in pygeoprocessing is wrong -- in fact i think it was wrong before when it was defining a distance to a channel that wasn't there.

But maybe those areas (that don't drain to any stream) _should_ be nodata. I'll discuss with the rest of the software team tomorrow.

> I don't quite understand this. Would this happen when there's NoData in the DEM, or something like a big pit that isn't filled by the model? If so, I'd suggest that they should check out their DEM for missing data or incorrect pits. What would using larger input data do?

I was thinking this would happen if the DEM didn't completely cover the watershed, where an area should drain to a stream, but doesn't because the stream is off the map.

This is my biggest concern with adding the pour points - it's difficult to explain and could further confuse users and ourselves) On the other hand, it might be clearer where the problem in the DEM is.

emlys:
@newtpatrol and @jagoldstein, this is ready for another look-over whenever you have time! We are trying out an integrated ReadTheDocs build that lets you see the built docs with latest changes right in the browser. In the Checks box at bottom of this page, click "Show all checks", then click on "Details" for the **docs/readthedocs.org:invest-userguide** check.

After discussion with Rafa, it seemed best to remove the whole pour points thing and let the nodata areas propagate through the outputs. I have updated the SDR and NDR sections to explain that. Thanks Stacie for questioning that approach!

The changes to the equations in SDR for _Rᵢ_ and _Fᵢ_ are to clarify that the _E'ᵢ_ term is outside the summation. This is how it is in the code and I confirmed with Rich that it's correct.

source/ndr.rst
```diff
@@ -183,6 +185,24 @@ Total nutrient at the outlet of each user-defined watershed is the sum of the co
...
 
 
+Defined Area of Outputs
+^^^^^^^^^^^^^^^^^^^^^^^
+
+NDR and several other model outputs are defined in terms of distance to stream (:math:`d_i`). Therefore, these outputs are only defined for pixels that drain to a stream on the map (and so are within the streams' watershed). Pixels that do not drain to any stream will have nodata in these outputs. Increasing the threshold flow accumulation value may shrink the defined area of these outputs, as fewer stream pixels exist and the watershed is smaller. Correspondingly, decreasing the threshold flow accumulation value may expand the defined area of these outputs.
+
+The affected output files are: **d_dn.tif**, **dist_to_channel.tif**, **ic_factor.tif**, **ndr_n.tif**, **ndr_p.tif**, **sub_ndr_n.tif**, **sub_ndr_p.tif**, **n_export.tif**, and **p_export.tif**.
+
+The model's stream map (**stream.tif**) is calculated by thresholding the flow accumulation raster (**flow_accumulation.tif**) by the threshold flow accumulation value:
```
jagoldstein, line 195:
```suggestion
The model's stream map (**stream.tif**) is calculated by thresholding the flow accumulation raster (**flow_accumulation.tif**) by the threshold flow accumulation value (TFA):

```

source/ndr.rst
```diff
@@ -183,6 +185,24 @@ Total nutrient at the outlet of each user-defined watershed is the sum of the co
...
+
+  .. math::
+     :label: ndr_stream
+
+     stream_{TFA,i} = \left\{\begin{array}{lr}
+          1, & \text{if } flow\_accum_{i} \geq TFA \\
+          0,     & \text{otherwise} \\
+          \end{array}\right\}
+
+
```
jagoldstein, line 205:
```suggestion

```

source/ndr.rst
```diff
@@ -327,7 +347,7 @@ Comparison to observed data
...
 
-If there are dams on streams in the analysis area, it is possible that they are retaining nutrient, such that it will not arrive at the outlet of the study area. In this case, it may be useful to adjust for this retention when comparing model results with observed data. For an example of how this was done for a study in the northeast U.S., see Griffin et al 2020. The dam retention methodology is described in the paper's Appendix, and requires knowing the nutrient trapping efficiency of the dam(s). 
+If there are dams on streams in the analysis area, it is possible that they are retaining nutrient, such that it will not arrive at the outlet of the study area. In this case, it may be useful to adjust for this retention when comparing model results with observed data. For an example of how this was done for a study in the northeast U.S., see Griffin et al 2020. The dam retention methodology is described in the paper's Appendix, and requires knowing the nutrient trapping efficiency of the dam(s).
```
jagoldstein, line 351:
```suggestion
```

source/sdr.rst
```diff
@@ -81,11 +90,19 @@ To avoid overestimation of the LS factor in heterogeneous landscapes, long slope
...
+   m &=  \left\{\begin{array}{lr}
+      0.2, & \text{where } \theta \leq 1\% \\
+      0.3, & \text{where } 1\% < \theta \leq 3.5\% \\
+      0.4, & \text{where } 3.5\% < \theta \leq 5\% \\
+      0.5, & \text{where } 5\% < \theta \leq 9\% \\
+      \beta / (1 + \beta), & \text{where } \theta > 9\%
+   \end{array}\right\} \\
+   \\
+   \beta &= \frac{\sin\theta / 0.0896}{3\sin\theta^{0.8} + 0.56}
+   \end{align*}
```
jagoldstein, line 106:
```suggestion

```

source/sdr.rst
```diff
@@ -181,12 +240,12 @@ The :math:`d` in :math:`dR_i` indicates a delta difference and :math:`p(i,k)` is
```
jagoldstein, line 240:
```suggestion

```

source/sdr.rst
```diff
@@ -195,6 +254,33 @@ Optional Drainage Layer
 In some situations, the index of connectivity defined by topography does not represent actual flow paths, which may be influenced by artificial connectivity instead. For example, sediments in urban areas or near roads are likely to be conveyed to the stream with little retention. The (optional) drainage raster identifies the pixels that are artificially connected to the stream, irrespective of their geographic position (e.g. their distance to the stream network). Pixels from the drainage layer are treated similarly to pixels of the stream network; in other words, the downstream flow path will stop at pixels of the drainage layer (and the corresponding sediment load will be added to the total sediment export).
```
jagoldstein, line 256:
```suggestion

```

source/sdr.rst
```diff
@@ -195,6 +254,33 @@ Optional Drainage Layer
...
 
+Defined Area of Outputs
+^^^^^^^^^^^^^^^^^^^^^^^
+
+SDR and several other model outputs are defined in terms of distance to stream (:math:`d_i`). Therefore, these outputs are only defined for pixels that drain to a stream on the map (and so are within the streams' watershed). Pixels that do not drain to any stream will have nodata in these outputs. Increasing the threshold flow accumulation value may shrink the defined area of these outputs, as fewer stream pixels exist and the watershed is smaller. Correspondingly, decreasing the threshold flow accumulation value may expand the defined area of these outputs.
+
+The affected output files are: **d_dn.tif**, **ic.tif**, **e_prime.tif**, **sdr_factor.tif**, **sdr_bare_soil.tif**, **d_dn_bare_soil.tif**, **ic_bare_soil.tif**, **sed_retention.tif**. **sed_retention_index.tif**, **sediment_deposition.tif**, and **sed_export.tif**
+
+The model's stream map is the union of the calculated stream layer and the input drainage layer (if provided).
+The model calculates a stream layer (**stream.tif**) by thresholding the flow accumulation raster (**flow_accumulation.tif**) by the threshold flow accumulation value:
```
jagoldstein, line 265:
```suggestion
The model calculates a stream layer (**stream.tif**) by thresholding the flow accumulation raster (**flow_accumulation.tif**) by the threshold flow accumulation value (TFA):

```

source/sdr.rst
```diff
@@ -195,6 +254,33 @@ Optional Drainage Layer
...
+          0,     & \text{otherwise} \\
+          \end{array}\right\}
+
+If the optional drainage input is provided, the model includes it (**stream_and_drainage.tif**):
+
+  .. math:: stream_{drainage,i} = stream_{TFA,i} \text{  OR  } stream_{input,i}
+     :label: stream_and_drainage
+
+The final stream layer (:math:`stream_{TFA}`, or :math:`stream_{drainage}` if the optional drainage input is provided) is used to determine :math:`d_i` for the SDR calculations.
+
```
jagoldstein, line 281:
```suggestion

```

source/sdr.rst
```diff
@@ -195,6 +254,33 @@ Optional Drainage Layer
...
+
```
jagoldstein, line 282:
```suggestion

```

jagoldstein (commented):
Thanks for making these changes @emlys . They mostly LGTM. I am good with removing the pour points and with the changes to the equations. My nit-picky two cents are that the "TFA" abbreviation should be established the first time "threshold flow accumulation" appears in each chapter, and especially before it appears in an equation. I am ok with spelling out "threshold flow accumulation" again later in the subsection that explains what it is, since that's the purpose of that part.
All the extra line breaks bother me, but I'm unsure if they actually affect landing page rendering. I'd prefer to stick with only a single line between paragraphs, per our [Style Guide](https://docs.google.com/document/d/1BHwHDu_I-x0s_2GsbUb4rfVmXMkl7kl97sx2suBTLh8/edit?usp=sharing).

Perhaps most importantly, some of the referenced equations numbers in the NDR section are inaccurate. For example, look at the intermediate outputs subsection of interpreting results. "effective_retention_x" is defined by equation 36, not 6, and there are at least two other intermediate outputs with the same issue. The links are also missing. I don't want to attempt inline code suggestions for this because I'm afraid I'll do it wrong and break something. I'm not requesting changes officially here, but please do update these equations. Thanks again!

jagoldstein (approved):

emlys:
Thanks @jagoldstein. I applied your commit suggestions and fixed the NDR equations. I'm going to hold off merging this until we do the 3.9.1 release.

> All the extra line breaks bother me, but I'm unsure if they actually affect landing page rendering. 

They do not affect the rendering. In RST, one line break in the source has no effect on the output. Two or more line breaks in the source make a new paragraph in the output. There may be some rare occasion where three breaks are needed to escape a directive or something. If you want additional line breaks to render, put a `|` at the start of the lines.
I'm fine with removing the breaks you suggested, but I do sometimes find them helpful as a visual separation when working on the source RST.

> some of the referenced equations numbers in the NDR section are inaccurate. 

This is unfortunately a problem throughout the user's guide. All references to equations should be done with `:eq:` to prevent getting out-of-order like this. Also, equations are numbered relative to the entire UG, so they start counting at a different number in each UG page. I fixed the NDR references but this is a bigger issue that will need to be addressed in the future (see #16 )

source/sdr.rst
```diff
@@ -181,19 +239,44 @@ The :math:`d` in :math:`dR_i` indicates a delta difference and :math:`p(i,k)` is
...
 
+Defined Area of Outputs
+^^^^^^^^^^^^^^^^^^^^^^^
+
+SDR and several other model outputs are defined in terms of distance to stream (:math:`d_i`). Therefore, these outputs are only defined for pixels that drain to a stream on the map (and so are within the streams' watershed). Pixels that do not drain to any stream will have nodata in these outputs. Increasing the threshold flow accumulation value may shrink the defined area of these outputs, as fewer stream pixels exist and the watershed is smaller. Correspondingly, decreasing the threshold flow accumulation value may expand the defined area of these outputs.
```
newtpatrol, line 258: This is very confusing to me. I have never seen increasing the TFA shrink the output area, it just creates fewer tributaries. Increasing the TFA has never expanded the area, it just creates more tributaries. If it did "expand", it couldn't go past the boundary of the input data. I'm guessing that this verbiage is related to removing the pour points? If so, instead of the shrink/expand text that's here, I'd suggest further explaining why they might not drain to a stream (going off the map, or pits/errors in the DEM), and describing what the output will look like for pixels that do not drain to the stream (you'll see areas of NoData that can't be explained by missing data in your input rasters...) While I'm thinking about this, was it crazy to suggest that these pixels receive a distinctly bogus value like -9999, so they'd be really obvious to people, and they wouldn't have to pour through their inputs to see what's causing NoData in the output?

newtpatrol (changes requested):
Aside from the text about TFA growing/shrinking output area, the rest of it looks fine to me at the moment.

source/sdr.rst
emlys: Yes, this is new behavior related to removing the pour points. I tried to revise the text as you suggest and added an image comparing the stream and SDR outputs with different TFA values. I'm struggling to explain this effectively.

Having a separate "nodata" value like -9999 could definitely be useful, and I made a ticket for it: #621 However it would require more not-insignificant code changes. 3.9.1 is just a bugfix release and in the interest of getting it out, I think we should come back to that idea as a possible feature in 3.10 or so.

newtpatrol: Thanks for working on this @emlys!  I'm about to take off for the day, but will take a look at the graphics, and would be happy to help word the explanation.

And thanks for considering the value-other-than-NoData for disconnected pixels. No hurry, just an idea.

emlys: Thanks @newtpatrol!

emlys:
I'm going to merge this now since the 3.9.1 release is imminent and seems like we resolved most if not all the discussions. Happy to continue improving these docs in future PRs.

PR #39, opened 2021-08-10
Configure RTD and GHA builds for natcap.invest dependency

### LINKED ISSUE #38: Easier way to view built UG artifacts
Currently, the only way to view the UG artifact uploaded from our check workflow is to navigate to the Checks > Artifacts menu, download a zip file, unzip it, and open one of the files in your browser locally.

Now that user's guide development is becoming more complex and folks outside of the software team are contributing a lot, it would be neat to have an easier way to see how our changes render.

We could start by activating the "Upload to GCS" step for PRs. With the ordinary GHA workflow, we could print out a link to the uploaded index.html page. With the [Checks API](https://docs.github.com/en/rest/reference/checks), I believe we could make a custom check that displays in the checks section at the bottom of the "Conversation" tab, and its "Details" could link to the uploaded index.html. That would probably be the most streamlined and conducive to us actually looking at it.

emlys:
Configure ReadTheDocs build and Github Actions build to work with the new `natcap.invest` dependency.
Fixes #38.
This isn't really needed until #34 is merged, but it seemed easier to do in a separate PR.

Since `pip install natcap.invest` does not work if `gdal` and some other dependencies aren't already installed, this is more complicated than just adding `natcap.invest` to the requirements.txt.

There are 2 non-trivial things needed to get a working RTD build:
- Checking out `invest-sample-data`, which some of the rst files depend on
- Installing `natcap.invest`, which is needed for the custom sphinx extension.

The `.readthedocs.yml` config is pretty limited in what you can do. The sphinx config, `source/conf.py`, which is executed on `sphinx-build`, is the only place we can put arbitrary commands, and so the only place we can check out `invest-sample-data`.

The RTD config allows you to install dependencies from conda, pip, APT, or `setup.py install`. `gdal` can only be installed with `conda`, and `natcap.invest` can only be installed with `pip`, so we need both.


source/conf.py
```diff
@@ -7,6 +8,10 @@
 # sys.path.append(os.path.abspath('.'))
 
 # -- General configuration -----------------------------------------------------
+if not os.path.exists('../invest-sample-data'):
+    subprocess.run(['make', '-C', '..', 'get_sampledata'])
+if not os.path.exists('invest-sample-data/pollination/landcover_biophysical_table_modified.csv'):
```
emlys, line 13: This isn't very neat but I think for now it is okay, we will probably not keep building on GHA and RTD simultaneously for a long time.

environment.yml
```diff
@@ -0,0 +1,12 @@
+channels:
+- conda-forge
+# set nodefaults to remove the default channels
+# this is needed to keep memory use below RTD limit due to conda bug:
+# https://github.com/conda/conda/issues/5003
```
davemfish, line 5: Wow, that issue is kind of a big deal. Good to know about this workaround!

requirements.txt
```diff
@@ -1,6 +1,4 @@
-# These are the packages that need to be installed in order to build HTML
-# documentation of the User's Guide.
-
+natcap.invest
```
davemfish, line 1: Since we tend to update the User's Guide `main` branch anytime we make invest changes on `main` (if the changes UG edits) does it make sense to install the latest invest `main` from github instead of the latest release on pypi? 

Stole this from our readthedocs: `pip install "git+https://github.com/natcap/invest@main#egg=natcap.invest"`

source/conf.py
davemfish: So this `conf.py` runs during both workflows, but with GHA we will have already done this prep in earlier workflow steps? 

Seems like a fine solution to me.

environment.yml
```diff
@@ -0,0 +1,12 @@
...
+- nodefaults
+dependencies:
+- pip
+- gdal==3.1.2
```
davemfish, line 9: I can imagine a time where we shift the gdal requirement for invest but forget to do it here and pip ends up trying to install a compatible gdal during `pip install natcap.invest`, and fails of course.

I'm not quite sure how we avoid that though...

davemfish (commented):
@emlys it's awesome to be able to click that readthedocs link and view the UG! I'd say that proves this all works well. I had a few questions/comments to consider before merging.

requirements.txt
emlys: Sure! In the longer term I wonder if we might start keeping separate "stable" and "latest" versions of the UG.

source/conf.py
emlys: Yea, this is just for RTD

environment.yml
emlys: That's a good point. I can't think of a way around it though... you could `curl` https://github.com/natcap/invest/blob/main/requirements.txt and get the latest gdal requirement, but on RTD there's no way to do that before starting the build.

emlys: I guess you could still do it in `conf.py`, but that's really supposed to be config for Sphinx, not general build setup.

davemfish (approved):
Okay I'll go ahead and merge. When it comes time to merge the other PRs that actually make use of the `natcap.invest` dependency, we can decide if there are more preferred ways to set the versions of `natcap.invest` and `gdal`.

PR #47, opened 2021-09-10
Consolidate text between UG and args spec (part 3 of 3)

emlys:
- Recreation
- RouteDEM
- Scenario Generator
- Scenic Quality
- SDR
- SWY
- Urban Cooling
- Urban Flood Risk
- Wave Energy
- Wind Energy


source/recreation.rst
```diff
@@ -7,7 +7,7 @@ Visitation: Recreation and Tourism
...
+Recreation and tourism are important components of many national and local economies and they contribute in innumerable ways to quality of life, sense of place, social connection, physical wellbeing, learning, and other intangibles. To quantify the value of natural environments, the InVEST recreation model predicts the spread of person-days of recreation, based on the locations of natural habitats and other features that factor into people's decisions about where to recreate. The tool estimates the contribution of each attribute to visitation rate in a simple linear regression. In the absence of empirical data on visitation, we parameterize the model using a proxy for visitation: geotagged photographs posted to the website flickr. Using photo-user-day estimates, the model predicts how future changes to natural features will alter visitation rates. The tool outputs maps showing current patterns of recreational use and maps of future patterns of use under alternate scenarios.
```
dcdenu4, line 10: Just checking that the move to single space vs double space for sentences was a choice and not an editor preference change.

source/recreation.rst
```diff
@@ -70,43 +70,32 @@ If a regression is computed, the model can optionally estimate visitation rates
...
 
-1. **Workspace (required).** Users must specify a path to the workspace folder where the tool will save its results::
+- :investspec:`recreation.recmodel_client workspace_dir`
 
-     Name: Path to a workspace folder.  Avoid spaces.
-     Sample path: \InVEST\Recreation\
+- :investspec:`recreation.recmodel_client results_suffix`
 
-2. **Results Suffix (optional).** Optionally enter a text string that will be used as a suffix on all output filenames.
+- :investspec:`recreation.recmodel_client aoi_path`
```
dcdenu4, line 85: It's a bit subjective but I wonder if some of that helpful text should end up in the "about" section of the spec or just added in on over here. Maybe:

>It is recommended that this shapefile be projected into linear units, especially if it is used to calculate a regression and scenario. Results are aggregated to these polygons.

source/scenario_gen_proximity.rst
```diff
@@ -142,112 +142,62 @@ generator is a base land-use/land-cover map and user-defined land cover
...
 Interpreting Results
 --------------------
 
 Final Results
 ~~~~~~~~~~~~~
 
--  **InVEST….log…txt:** Each time the model is run, a text (.txt) file
-   will appear in the \ *Output* folder. The file will list the
+- **InVEST….log…txt:** Each time the model is run, a text (.txt) file
+   will appear in the *Output* folder. The file will list the
```
dcdenu4, line 174: Looking at the html there's a newline at the end of ` a text (.txt) file` that breaks up the text. It's not a big deal but occurs for the rest of the Final and Intermediate results.

source/recreation.rst
emlys: Yeah, I did it on purpose. A lot of files had some double spaces mixed in, but mostly used single space. Super minor but it gets annoying once you start noticing it!

source/scenario_gen_proximity.rst
emlys: good catch!

source/recreation.rst
emlys: Added this back in. Should it be marked as `"projected": True` in the args spec? I think I took this out originally because it wasn't. But we don't have to decide now

dcdenu4: Yea, I'm not certain. But like you say, can handle that in the other related PR. I'm sure Dave will have some insight there.

dcdenu4 (approved):
Thanks @emlys! There's a conflict with requrements.txt but otherwise looks good to me. I'll go ahead and approve it now and you can merge after the conflict resolution!

emlys:
Thanks @dcdenu4! Since natcap/invest#639 was merged, I could point this to release/3.10 instead of the feature branch. Merging now!

PR #50, opened 2021-11-16
Internationalization

emlys:


README.md
```diff
@@ -26,3 +26,39 @@ For anything not listed in our style guide, follow the [Google developer documen
...
+   ```
+   Or to update PO files for all languages:
+   ```
+   $ sphinx-intl update --locale-dir source/locales --pot-dir build/gettext
+   ```
+3. Send the PO files in `source/locales/<LANG>/LC_MESSAGES` to the translator for the language. The translator will fill in the translations and send them back.
+4. Replace the PO files in `source/locales/<LANG>/LC_MESSAGES` with the updated ones from the translator.
+
+## text provided by Sphinx
+Sphinx comes with translations for messages it generates in several languages including Spanish and Chinese. Setting the [`language`](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-language)option will tell Sphinx to use messages in that language.
```
dcdenu4, line 51: Cool!

README.md
```diff
@@ -26,3 +26,39 @@ For anything not listed in our style guide, follow the [Google developer documen
...
+### To add support for a new language
+```
+$ sphinx-intl update --locale-dir source/locales --pot-dir build/gettext --language <LANG>
+```
+This will create PO files in `source/locales/<LANG>/LC_MESSAGES/` copied from the POT files in `build/gettext`.
+
+The language will most likely be on the ([list of languages supported by Sphinx](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-language)). But if not, we'll need to figure out how to provide our own message catalog for Sphinx-generated messages.
+
+## To build docs in a different language
+
```
dcdenu4, line 64: Should this be filled in or is it leftover?

source/conf.py
```diff
@@ -14,7 +14,11 @@
 
 # Add any Sphinx extension module names here, as strings. They can be extensions
 # coming with Sphinx (named 'sphinx.ext.*') or your custom ones.
-extensions = ['sphinx.ext.mathjax', 'investspec']
+extensions = [
+    'sphinx.ext.mathjax',
+    'sphinx_reredirects',
```
dcdenu4, line 19: Do we need `sphinx_intl` here as well?

source/locales/es/LC_MESSAGES/carbon_edge.po
```diff
@@ -0,0 +1,865 @@
+# SOME DESCRIPTIVE TITLE.
+# Copyright (C) 2021, The Natural Capital Project
+# This file is distributed under the same license as the InVEST package.
+# FIRST AUTHOR <EMAIL@ADDRESS>, 2021.
+#
+#, fuzzy
```
dcdenu4, line 6: Out of curiosity, what does the `, fuzzy` indicate?

source/locales/es/LC_MESSAGES/the_need_for.po
```diff
@@ -0,0 +1,473 @@
...
+"ecosystem services and the benefits provided to people"
+msgstr ""
+
+#: ../../source/the_need_for.rst:45
+msgid ""
+"“Supply” represents what is potentially available from the ecosystem (ie."
+" what the ecosystem structure and function can provide). For example, "
+"this would be the wave attenuation and subsequent reduction in erosion "
+"and flooding onshore provided by a particular location and density of "
+"mangrove forest. “Service” incorporates demand and thus uses information "
```
dcdenu4, line 175: If a translator goes to update the `msgstr` will they be able to enter escapable (?) quotes like this?

source/locales/es/LC_MESSAGES/the_need_for.po
```diff
@@ -0,0 +1,473 @@
...
+"Capital Project has used InVEST in over 60 countries worldwide. See the "
+"Where We Work section of the NatCap website "
+"(https://naturalcapitalproject.stanford.edu/impact/where-we-work) for the"
+" latest map and description of our projects. Through our experience "
+"applying InVEST and helping to shape decisions, we have seen how the "
+"InVEST tool fits within the larger context of a natural capital approach."
+msgstr ""
+
+#: ../../source/the_need_for.rst:114
+#, python-format
```
dcdenu4, line 330: Out of curiosity, what's the `, python-format` here?

source/conf.py
emlys: No, `sphinx_intl` isn't a sphinx extension, just a command line tool.

emlys:
@dcdenu4, I finished the README and added a bit of functionality that was missing (to translate the handful of strings that live in `investspec.py`).

extensions/investspec/investspec.py
```diff
@@ -1,18 +1,19 @@
...
-    'LINESTRING',
-    'MULTILINESTRING',
-    'POLYGON',
-    'MULTIPOLYGON']
+    _('POINT'),
+    _('MULTIPOINT'),
+    _('LINESTRING'),
+    _('MULTILINESTRING'),
+    _('POLYGON'),
+    _('MULTIPOLYGON')]
```
dcdenu4, line 16: Ah, interesting. Certainly a lot to keep track of!

extensions/investspec/investspec.py
```diff
@@ -25,6 +26,23 @@ def format_type_string(arg_type):
     Returns:
         formatted string that links to a description of the input type(s)
     """
+    # some types need a more user-friendly name
+    # all types are listed here so that they can be marked up for translation
+    type_names = {
```
dcdenu4, line 31: I have a stupid question, are these types defined somewhere on the `natcap.invest` side? I looked around but didn't find a similarly valid types structure... I was mostly curious because I was wondering if we could take advantage of a single source of truth on the `natcap.invest` side in case types were added in the future. Not that'd we really be messing with these types all that much. Sorry about this ramble.

extensions/investspec/internationalization/messages.pot
```diff
@@ -0,0 +1,124 @@
+# Translations template for InVEST.
+# Copyright (C) 2021 Natural Capital Project
+# This file is distributed under the same license as the InVEST project.
+# FIRST AUTHOR <EMAIL@ADDRESS>, 2021.
+#
```
dcdenu4, line 5: Do we want both this file, `messages.pot` and the above file `messages.po`? I think I read in one of your Readmes that we can only need to check one in because the other is easily derived? Maybe it's different for this case? Sorry if I'm not fully understanding or confusing something!

source/locales/es/LC_MESSAGES/carbonstorage.po
```diff
@@ -0,0 +1,1232 @@
...
+msgstr ""
+
+#: ../../source/carbonstorage.rst:89
+msgid "C_dead"
+msgstr ""
+
+#: ../../source/carbonstorage.rst:91
+msgid "1"
+msgstr ""
+
```
dcdenu4, line 413: Interesting to see how tables get broken out.

source/locales/es/LC_MESSAGES/coastal_vulnerability.po
```diff
@@ -0,0 +1,1986 @@
...
+#: ../../source/coastal_vulnerability.rst:38
+msgid ""
+"The model calculates the exposure index and coastal population density "
+"using a spatial representation of the following bio-geophysical "
+"variables:"
+msgstr ""
+
+#: ../../source/coastal_vulnerability.rst:40
+#: ../../source/coastal_vulnerability.rst:108
+#: ../../source/coastal_vulnerability/ranking_system_table.csv:1
```
dcdenu4, line 171: Oh cool, it's smart enough to cut down on redundancy.

source/locales/es/LC_MESSAGES/contents.po
```diff
@@ -0,0 +1,19 @@
+# SOME DESCRIPTIVE TITLE.
+# Copyright (C) 2021, The Natural Capital Project
+# This file is distributed under the same license as the InVEST package.
+# FIRST AUTHOR <EMAIL@ADDRESS>, 2021.
```
dcdenu4, line 4: Is it possible to skip `contents.rst`? I don't think there's anything to translate.

extensions/investspec/internationalization/locales/es/LC_MESSAGES/messages.po
```diff
@@ -0,0 +1,125 @@
...
+# FIRST AUTHOR <EMAIL@ADDRESS>, 2021.
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: InVEST 3.10\n"
+"Report-Msgid-Bugs-To: esoth@stanford.edu\n"
+"POT-Creation-Date: 2021-11-30 13:20-0800\n"
+"PO-Revision-Date: 2021-11-30 13:21-0800\n"
+"Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
+"Language: es\n"
```
dcdenu4, line 13: I don't think I saw this key in the other PO files, should we update the other PO files for consistency?

dcdenu4 (changes requested):
Hey @emlys , this is really cool and was a great way to learn more about how translations is working across InVEST. I had a few questions / comments and one general question below.

Are we committing PO files for Spanish as demonstration of what would eventually be done and as a way to test the functionality of this PR?

source/locales/es/LC_MESSAGES/carbon_edge.po
emlys: It's put there by the tool that generates the PO file. [This post](https://stackoverflow.com/questions/15653093/what-does-fuzzy-mean-at-the-top-of-a-xgettext-generated-pot-file) suggests it's to indicate that the header needs to be filled in. I think that's okay for now since we haven't filled in the files yet.

extensions/investspec/investspec.py
emlys: I don't believe there is... it would make sense to import them from `natcap.invest`. `natcap.invest` itself has no use for _translations_ of the type names, but since they're used here and would be used in the workbench, we could centralize them there.

extensions/investspec/internationalization/messages.pot
emlys: I think you're thinking of the `messages.mo` that we don't need to include... those are binary "compiled" versions of the PO files. The POT and PO files should both be checked in because the POT file is the source of truth for the current set of translated strings, while the PO files store the translations for each language.

extensions/investspec/internationalization/locales/es/LC_MESSAGES/messages.po
emlys: I think it's because this PO file is created by pybabel while the others are created by sphinx-intl. I guess they have slightly different headers (all the header fields are optional according to the GNU gettext standard). I don't feel it really matters since the language is defined in the path but can add it to the other PO files if you prefer

source/locales/es/LC_MESSAGES/contents.po
emlys: `sphinx-intl` doesn't seem to have any option to skip a file - it makes one PO file for each RST file in a given directory. And this file will get re-created whenever we do `sphinx-intl update`. We could delete it now and each future time, but I think it's also okay to keep it here. No strong opinion.

source/locales/es/LC_MESSAGES/carbonstorage.po
emlys: Maybe I should go through and remove the tables from the PO files... `c_dead` for instance should not be translated.

extensions/investspec/investspec.py
emlys: Oh, yeah. I wish there weren't so many separate translation mechanisms going on! Maybe a simpler way will become clear once we have used it a bit.

source/locales/es/LC_MESSAGES/the_need_for.po
emlys: Oh interesting, I hadn't noticed that. The comment lines starting with `#,` are flags which have special meaning to gettext programs. The `python-format` flag indicates that the message is a python format string (clearly incorrect here). I suspect it's because of the `%` in the string. Though, it's strange that the sphinx gettext utility isn't correcting for that.

extensions/investspec/internationalization/locales/es/LC_MESSAGES/messages.po
dcdenu4: I don't think I'm familiar enough to have a preference! But I think you're right, language is embedded in the path so not necessary.

source/locales/es/LC_MESSAGES/contents.po
dcdenu4: I think it's fine to leave as is.

source/locales/es/LC_MESSAGES/carbonstorage.po
dcdenu4: I wish `sphinx-intl` had a configuration to allow for ignoring certain RST syntax, like tables. Removing table sections from the PO files each time they need to be updated seems fairly onerous. A quick google search didn't bring up anything related unfortunately.

extensions/investspec/investspec.py
dcdenu4: Okay, it's fine to leave as is for this PR then. I think it's fine to wait until this actually pops up as an issue before centralizing things in `natcap.invest`.

source/locales/es/LC_MESSAGES/carbonstorage.po
emlys: I wonder if it would be easier not to use `gettext` at all for the RST docs and try to get them translated directly by someone who's familiar with RST syntax.

source/locales/es/LC_MESSAGES/the_need_for.po
emlys: Pretty sure this is a bug in the sphinx gettext builder.

source/locales/es/LC_MESSAGES/carbonstorage.po
emlys: I made a ticket to revisit this in the future: #54

source/locales/es/LC_MESSAGES/the_need_for.po
emlys: Normal quotes can be escaped with a backslash: `\"`. These fancy quotes `“”` are used in the source RST. Is that what you meant?

emlys:
@dcdenu4, since we're not going to get translations done before the release, I don't think the PO files should be committed. It doesn't really matter here, but I'm considering the contents of `source/locales/` to be the single source of truth for which languages are supported. I removed all PO files in 04bbadc.

emlys:
This depends on the changes in natcap/invest#645 so let's merge this after that PR goes through.

dcdenu4 (commented):
Hey @emlys, thanks for the update and creating that well defined issue about using gettext here vs translating the RST directly. I agree with removing the PO files for now until we actually have a translation case, but it was great to see how it would work and what they'd look like!

Looks like there are some conflicting files and seeing how we're waiting on another PR I'll hold off on approving so we remember to wait for that other PR and handle the conflicts.

emlys:
@dcdenu4 natcap/invest#568 was merged, so this is ready now!

dcdenu4 (approved):
Thanks for all of this work and thought @emlys!

Issue #53, opened 2021-12-03
Document new mask raster in NDR/SDR


phargogh:
The mask raster we're producing in https://github.com/natcap/invest/issues/621 would be helpful to include in the SDR page's graphic illustrating the effects of different TFAs.

Currently working on this in my fork on `feature/ndr-sdr-what-drains-to-stream`.

phargogh:
@emlys and I were discussing this a little on the side, and it would be really nice to be able to have a script to automate the creation of this image, probably by running SDR or NDR, stylizing the outputs and using `matplotlib` or something similar to build up the PNG to then display in the UG chapter itself.

Although it might be cool to do so, because this involves running a model twice with different TFAs, this need not be run as part of every build.

emlys:
@phargogh, I'm happy to make the script for that

emlys:
@phargogh, did you have something written up for this already? Just asking because of a related forums question.

phargogh:
I do have a branch by that name, but there aren't any meaningful changes.  I think your matplotlib script in #56 superseded those changes.

emlys:
Cool, I just added mentions of `what_drains_to_stream.tif` to the SDR and NDR pages.

PR #56, opened 2021-12-10
Script and data to generate SDR TFA effects image

emlys:
Addresses part of #53. I also added a third column so it shows results for TFA of 100, 400, and 1000.


emlys:
@phargogh I've updated this with the new `what_drains_to_stream` rasters!

source/sdr/tfa_effects/README.md
```diff
@@ -0,0 +1,5 @@
+# script and data used to generate the image `tfa_effects.png`
+
+The image `tfa_effects.png` is meant to illustrate the effect of the TFA value on the stream, SDR, and mask outputs. It was generated using the script and data in this folder. The data are outputs from the SDR model run on the SDR test dataset. All parameters were set to the defaults, except TFA which was set to 100, 400, and 1000.
+
+Run `python tfa_effects_script.py` from within this directory to re-create the file `tfa_effects.png` (requires `matplotlib`, `numpy`, and `pygeoprocessing`).
```
phargogh, line 5: I see that the output TIFFs have been committed here, which is fine since the files themselves are so small, and it does make sense that we might take this approach since the test data isn't a dependency of the build process (and probably shouldn't be).

But I wonder, what would you think about either fleshing out the script to also run SDR the 3 times as is documented, or else creating a second script to run SDR to create these files?  Just thinking about future us and making it easier to update this when there are notable changes to the model.  On the other hand, the graphic is for illustrative purposes and maybe this isn't a useful thing for us to do at this point.

Anyways, what do you think about this?

phargogh (changes requested):
The script looks good to me, thanks for adding this!

My comments are mostly about when we're looking to reproduce these outputs in the future and the degree to which we should make it easier for future us at this stage.  I'm curious what you think!

source/sdr/tfa_effects/README.md
emlys: That makes sense! I updated the script to run SDR three times and so we don't need the TIFFs checked in. The final PNG is still checked in, so this won't have to run as part of the build process - just if we ever want to update it.

phargogh (approved):
Thanks!

source/sdr/tfa_effects/README.md
phargogh: Yep, I agree with this!  We don't want to have to rerun sdr on each UG build, so makes sense to commit the final graphic and we can update that committed graphic when we need to.

PR #60, opened 2022-01-12
Move arg formatting logic over to invest

emlys:
Related to natcap/invest#777. Goes with natcap/invest#814.
* Remove formatting logic
* Remove formatting tests
* Remove python gettext things
* Remove the `demo_investspec` make command
* Add integration tests for the investspec role

Lets merge natcap/invest#777 first, then I'll update `requirements.txt` to point to `natcap/invest@main` rather than the feature branch.


requirements.txt
```diff
@@ -1,5 +1,5 @@
 git-lfs
-natcap.invest
+git+https://github.com/emlys/invest.git@task/777
```
davemfish, line 2: I believe the changes are now in `natcap.invest`. Should we switch this back?

davemfish (changes requested):
Hi @emlys , I think the `requirements` can be switched back to `natcap.invest` here. Also the Actions never ran. But maybe they will when we push a new commit?

requirements.txt
emlys: They haven't been released yet, but I changed it to point to `natcap/invest@main`.

davemfish: Oh, right. I guess that's what I was thinking.

davemfish (approved):
I approve. Not sure what to do about the fact that Actions aren't running. But I guess RTD proves that things are working.

PR #69, opened 2022-06-09
trying to fix the investspec tests

emlys:
This PR fixes the github actions triggers, which were not triggering on every push and pull request as expected. It also corrects some accumulated errors that were not noticed before because the workflow didn't run.

### DIFF
--- a/.github/workflows/build-sphinx.yml
+++ b/.github/workflows/build-sphinx.yml
@@ -1,22 +1,9 @@
 name: User's Guide
 on:
-    schedule:
-        # Rebuild daily, just after (30 minutes after) any tags were updated.
-        # I'd prefer to rebuild in response to added tags, but that's
-        # much harder to make happen on github actions.
-        - cron: '30 0 * * *'
-
     push:  # Build the docs for any branch (we only push to GCP when building from main)
-        # branches:
-        #     - '*'
-
-    create:  # Any time a tag is added, rebuild the docs.
-        branches:
-            - main
-
     pull_request:  # Build the docs for any PR
-        branches:
-            - '*'
+    create:  # Any time a tag (or branch) is added, rebuild the docs.
+
 jobs:
     build-ug:
         name: Sphinx docs
--- a/.github/workflows/build-sphinx.yml
+++ b/.github/workflows/build-sphinx.yml
@@ -7,7 +7,7 @@ on:
 jobs:
     build-ug:
         name: Sphinx docs
-        runs-on: macos-10.15
+        runs-on: ubuntu-latest
         defaults:
           run:
             shell: bash -l {0}
--- a/.github/workflows/build-sphinx.yml
+++ b/.github/workflows/build-sphinx.yml
@@ -18,14 +18,14 @@ jobs:
             - name: Set up python
               uses: actions/setup-python@v1
               with:
-                  python-version: 3.10
+                  python-version: "3.10"
 
             - name: Setup conda environment
               uses: conda-incubator/setup-miniconda@v2
               with:
                   activate-environment: env
                   auto-update-conda: true
-                  python-version: 3.10
+                  python-version: "3.10"
                   channels: conda-forge
 
             - name: Install python dependencies with conda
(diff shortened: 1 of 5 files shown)


extensions/investspec/test/test_investspec.py
```diff
@@ -14,13 +14,13 @@
 class TestInvestSpec(unittest.TestCase):
 
     @classmethod
-    def beforeAll(cls):
+    def setUpClass(cls):
```
emlys, line 17: This function wasn't running because I named it wrong, `beforeAll` is the equivalent for Jest. 🤯

.github/workflows/build-sphinx.yml
```diff
@@ -1,26 +1,13 @@
...
-        # much harder to make happen on github actions.
-        - cron: '30 0 * * *'
-
     push:  # Build the docs for any branch (we only push to GCP when building from main)
-        branches:
-            - '*'
-
-    create:  # Any time a tag is added, rebuild the docs.
-        branches:
-            - main
```
emlys, line 15: I don't think this was having any effect because it's not listed in the [workflow syntax reference](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushbranchestagsbranches-ignoretags-ignore).

.github/workflows/build-sphinx.yml
```diff
@@ -1,26 +1,13 @@
...
 on:
-    schedule:
-        # Rebuild daily, just after (30 minutes after) any tags were updated.
-        # I'd prefer to rebuild in response to added tags, but that's
-        # much harder to make happen on github actions.
-        - cron: '30 0 * * *'
-
     push:  # Build the docs for any branch (we only push to GCP when building from main)
-        branches:
-            - '*'
```
emlys, line 11: This and the same thing for `pull_request:` were somehow having the opposite effect. The `push` and `pull_request` events are for any branch by default.

.github/workflows/build-sphinx.yml
```diff
@@ -1,26 +1,13 @@
 name: User's Guide
 on:
-    schedule:
-        # Rebuild daily, just after (30 minutes after) any tags were updated.
-        # I'd prefer to rebuild in response to added tags, but that's
-        # much harder to make happen on github actions.
-        - cron: '30 0 * * *'
-
```
emlys, line 8: I think this is redundant with the `create:` trigger?

.github/workflows/build-sphinx.yml
```diff
@@ -1,26 +1,13 @@
...
     pull_request:  # Build the docs for any PR
-        branches:
-            - '*'
+    create:  # Any time a tag (or branch) is added, rebuild the docs.
+
 jobs:
     build-ug:
         name: Sphinx docs
-        runs-on: macos-10.15
+        runs-on: ubuntu-latest
```
emlys, line 10: macos runners are the least available, using ubuntu where possible might slightly reduce our wait times.

.github/workflows/build-sphinx.yml
```diff
@@ -31,46 +18,22 @@ jobs:
...
                   activate-environment: env
                   auto-update-conda: true
-                  python-version: 3.8
+                  python-version: "3.10"
                   channels: conda-forge
 
             - name: Install python dependencies with conda
               run: conda env update --file environment.yml
 
-            - name: Check for unused files
```
emlys, line 47: I added this step after removing a bunch of unused files in this repo. Since we rarely add files I no longer think it's useful. And it would have to be updated slightly to work on ubuntu.

.github/workflows/build-sphinx.yml
```diff
@@ -31,46 +18,22 @@ jobs:
...
-                    fi
-                  fi
-                done
-                exit $exitcode
-
             - name: Run custom extension tests
               run: make test_investspec
 
-            - name: Demo the custom extension
-              run: make demo_investspec
```
emlys, line 72: This target no longer exists.

phargogh (approved):
Thanks @emlys !  I particularly like the reworking of the build triggers to be event-based rather than based on a schedule, and the switch to ubuntu for builds makes a whole lot of sense.

phargogh:
We're still not sure what's going on with the RTD builds, but that's less important than getting the regular docs to build.

PR #74, opened 2022-07-25
Release/3.11.0

newtpatrol:
Significant updates throughout several sections of the SDR chapter, including:

- Added a new "Recent Changes" section at the beginning of the chapter.
- Updated R and T equations and related text.
- Changed instances of "deposit" to "trap".
- Added an Ecosystem Service Indicators section to The Model defining equations for new outputs.
- Added new text to the Sediment Retention Services section.
- Removed The Model section about Bare Soil calculations.

More work needs to be done by others, so merging this so everyone else can pull this version and work off of it.


source/sdr.rst
```diff
@@ -356,16 +364,24 @@ Translating the biophysical impacts of altered sediment delivery to human well-b
...
 
-Evaluating service entails locating the relevant beneficiaries on the landscape and linking them to sediment deposition (or change in sediment export). As an example for point beneficiaries such as a drinking water withdrawal, one method is to create the watershed that drains to that point location (using a tool like DelineateIt) and then sum sediment deposition (or change in sediment export) within that watershed.
+Evaluating service entails locating the relevant beneficiaries on the landscape and linking them to sediment trapping (or change in sediment export). As an example for point beneficiaries such as a drinking water withdrawal, one method is to create the watershed that drains to that point location (using a tool like DelineateIt) and then sum the avoided export output raster (or the change in sediment export, if working with scenarios) within that watershed.
+
+Example use cases for valuing sediment retention
+------------------------------------------------
+
+NOT SURE IF THIS IS THE RIGHT PLACE FOR THIS
+
+ADRIAN, LISA (and others?) PLEASE ADD A COUPLE OF EXAMPLES OF USING THESE OUTPUTS.
```
lmandle, line 374: @newtpatrol I don't have any examples yet with the new formulation -- though we should soon with the Colombia work. Do you think it would make sense to include the Myanmar example in the meantime, while noting that it was done with the old formulation of the model, so the specific outputs will differ? Or is that just confusing, and I should add the Colombia work with the new model when it's available? What do you think would be the most helpful/least confusing to InVEST users?

lmandle (commented):
This looks great, Stacie! 

See my comment for what is really a question about how to make the examples most useful.

source/sdr.rst
newtpatrol: Thanks @lmandle. I also realized that we don't have examples with the new results, then forgot to flag it to others, so thanks. I believe that @adrianvogl proposed adding this section, so perhaps she could weigh in on whether to give old examples, or perhaps just nix this section for now and add it later, when we have updated examples to provide?

PR #85, opened 2022-08-24
Release/3.12.0

newtpatrol:
Not quite sure why I need to do a pull request for adding new versions of Rafa's PPT and PNG, but ok.


phargogh:
The new versions appear to have already been added in d9dc0de (which is on `release/3.12.0`) so I'll close this PR since we don't need it.

PR #86, opened 2022-09-01
Merge the `release/3.12` branch into `main` post-release

phargogh:
Just a simple merge here post-release!


requirements.txt
```diff
@@ -1,5 +1,6 @@
 git-lfs
-natcap.invest
+#natcap.invest
+git+https://github.com/natcap/invest.git@release/3.12#egg=natcap.invest
```
dcdenu4, line 3: Hey @phargogh, I don't think it makes sense to leave this in here right?

dcdenu4 (commented):
Hey @phargogh thanks for this. Just one question about the version of `invest` in the requirements file.

requirements.txt
phargogh: Ooo good catch!  Yes, I updated it in https://github.com/natcap/invest.users-guide/commit/30464f4c6f73feebd1b8a02b8e0800b9854ce89d to use `main` instead.  We had previously been using the latest released version of InVEST, but I think `main` might be a better fit.

phargogh:
Thanks @dcdenu4 !  I corrected this to build against `main` instead.  Let me know what you think!

PR #99, opened 2022-11-04
Update citation and contributors

dcdenu4:
This PR updates the contributors and citation in the UG. This work was a joint effort from the SHLT and lead by Adrian. 

- Removed acknowledgment section
- Added `custom.css` to be able to display contributors without bullets in list format. I created the separate CSS to better be able to save this functionality if we change themes as opposed to adding it into an existing CSS file.


phargogh (approved):
Looks good to me!  I also downloaded the built UG and verified that the custom CSS will only apply to the front page list and not to other bullet lists in the UG.

I see this is still in draft @dcdenu4 but I'll add my review anyways.  Thanks!

dcdenu4:
Thanks James! I remove Lisa as reviewer since I already had feedback from her via Slack, and with readthedocs not working it's harder to get a quick glance at the rendered UG result.

PR #101, opened 2022-11-16
Fix RTD build

dcdenu4:


emlys:
@dcdenu4 I think we can close this now?

dcdenu4:
Yep!

PR #102, opened 2022-11-30
Spanish translation

emlys:
To support multiple languages, this PR introduces a directory for each language:
* Move the English source docs into `source/en`
* Add all the Spanish source docs into `source/es`

They all share the same `conf.py` file. 

Changes to the build process:
* `make html` now builds each language as an independent sphinx project. 
* `make redirect` builds another copy of the docs where each page redirects to the new location of that page in English (ie. `build/html/foo.html -> build/html/en/foo.html`). That way we won't break links that point to the old user's guide location.
* The github actions workflow now uploads dev builds of the user's guide to the dev builds bucket. 
    * Like with invest, this doesn't work on PRs - only on your fork. You'll need to add the GCP authentication secret as well.
    * Because we don't version the user's guide, only the "latest" dev version is hosted for each user. This is the version most recently built on your fork.
    * For example: https://storage.googleapis.com/natcap-dev-build-artifacts/invest-userguide/emlys/latest/es/index.html


source/es/index.rst
```diff
@@ -0,0 +1,177 @@
...
+   * Zheng, Hua
+   * Ziv, Guy
+
+Además, agradecemos la ayuda de cientos de colaboradores co-creando
+soluciones en compromisos en todo el mundo.
+
+\*Hacemos todo lo posible para mantener esta lista completa y actualizada, pero reconocemos
+es posible que hayamos perdido a alguien sin darnos cuenta. Si no ve su nombre en la lista
+anterior y cree que debería serlo, comuníquese con
+naturalcapitalproject@stanford.edu.
```
emlys, line 111: I google translated this, I hope that's okay!

phargogh: Yep, seems fine to me!

emlys:
Thanks @phargogh! I also had to remove the reference to `ws_id` in SDR.

PR #108, opened 2023-01-25
Add Urban Nature Access Chapter

phargogh:
For the ongoing edits and revisions of the Urban Nature Access UG chapter.


newtpatrol:
Here are some initial questions and suggestions about the UG chapter.

The Model
------------

This section would benefit from having conceptual diagrams that illustrate how the model works. Definitely at the beginning of The Model, perhaps elsewhere, like Two-Step Floating Catchment Area, and the different ways of running the model (with population radii etc).

"Decay Function" needs a sentence that explicitly says what "dichotomy", "exponential" etc are. As in (something like) "The following decay functions are provided as user-selectable model inputs:" 
- Also provide a brief explanation of Exponential, Power, etc, like is given for Dichotomy.

Is the "default model" the same as the "core model"? If so, please be consistent with which term is used throughout the chapter.

Recommend "Calculate Greenspace Supply" instead of "Two-Step Floating Catchment Area" to be more understandable and consistent with the following sections.

Typeo in Calculate Per-Capita Greenspace Demand - "foal" should probably be "goal"


Data Needs
-------------

At the beginning of Data Needs we should Note that sample data is supplied to provide an example of requirements and formatting. (This is probably true for other model chapters as well).
- Also, in other chapters we Note that layers will be resampled to match the DEM/LULC/whatever. If that's the case here, should be noted at the beginning. 

I also feel like we need to define up front what it means for an input to be "conditionally required". And/or, further define those conditions in each input description. Some entries already do this (like Power Function Beta Parameter). Others don't (like Population Group Radii Table).

Land use/land cover: Suggest adding a note that the name of the LULC file will be used for a gpkg output layer name. But... What happens if the LULC file name is really long, has spaces, other odd characters, etc? Do people need to be given rules for naming? Or could this output layer be named something standard and not tied to the complete unknown that is a user input file name?

Population Raster units - should they be "number of people" instead of "unitless"?

Administrative boundaries field: This needs more explanation. Where does POP_GROUP come from? Is there a naming requirement? Is it required to match something else? Provide an example (which might match the example needed in Population Group Radii Table?)

"Decay Function" and "Search radius mode" have some funky formatting. Other places too within the output listing etc, which I assume will get fixed in final editing.

Aggregate By Population Groups - change "user-defined admin units vector" to user-defined Administrative Boundaries vector" for consistency.

Population Group Radii Table - in pop_group, change "AOI vector" to what? Is "Administrative Boundaries" the AOI vector? Again, make the terms consistent throughout.
- Give an example of matching the pop_group field names.


Interpreting Results
--------------------

These all need units defined, and links back to the relevant equations to be consistent with other model chapters.

output/aois.gpkg - Again, "AOI" is not an input. Which input is the AOI? "Administrative Boundaries"? Use that term, or change the UI and UG to use the term Area of Interest (AOI). If it is specifically intended to be admin boundaries (which the model description indicates is true), then stick with that term and use it consistently between text, file naming and UI. 

SUP_DEMadm_cap - Instead of "greenspace budget" I'd use words that match what's in "SUP_DEMadm_cap", which contains none of "average greenspace budget". I know I'm harping on this, but consistency within the text and between text and file naming makes understanding so much easier. (And that field name could be improved anyway).

intermediate/aligned_lulc.tif - I'd say something like "A copy of the user’s land use land cover raster. If the user-supplied LULC has non-square pixels, they will be resampled to square pixels."

Not having really used the model, I do wonder why so many potentially interesting output layers are kept in the intermediate folder.

Something I think we should do more of in general is provide guidance about how one would use the different outputs. We added this to SDR in the recent renovation. This chapter provides none at all.

intermediate/greenspace_budget.tif is output by the model but not described in the UG. 
intermediate/reprojected_aois.gpkg is output by the model but not described in the UG.

I assume that a Data Sources Appendix will be added, like other chapters have?

phargogh:
Thanks for your feedback on this @newtpatrol !

I have updated the user's guide to add a few things that I feel confident about adding:

1. **Search Kernels**: Graphics demonstrating how the search kernels decay out to the search distance and beyond, along with a few descriptions where text made sense to add further context to the mathematical definition and the graph.
2. **default model**: renamed to "core model".
3. **"Two-Step Floating Catchment Area (2SFCA)"** section renamed to "Calculating Greenspace Supply"
4. **Note about sample data** added, along with a note about the target resolution and spatial projection.
5. **Conditionally required** context added in the model's ARGS_SPEC about the conditions under which the input, column or field is required.  This may eventually be handled more resiliently by the model spec, but this should be enough for now.
6. **Population raster units** corrected to "count".
7. **Administrative Boundaries** is now consistent throughout (or should be)
8. **Kernel file** removed from the description of model outputs.
9. **Units** some have been added, some more need to be added.  A TODO is up at the top to note the need.
10. **Various typos fixed**

A few notes about other things:

* This model has a _lot_ of summations and, depending on the search radius mode, different ways of slicing.  Because of this, it's difficult for me to conceive of how one might visually represent this.  I've added the 2SFCA graphics from the PEP, but I realize that's not a complete solution.  Got any ideas about how to visualize this?
* I'm waiting on the science team's input for a data sources appendix.
* Formatting of options (such as the kernel choice and search radius mode) is an issue thoughout the UG wherever there's a dropdown menu.  See https://github.com/natcap/invest/issues/1163
* It's easy to move files around and was not clear from the PEP which files should go where.  If you have recommendations about other outputs, let me know!

newtpatrol:
The search kernel graphs are very useful, thanks @phargogh!

I also see the Calculating Greenspace Supply graphics, and understand that things like this are hard to make visuals for. The ones from the PEP are indeed not optimal, at a minimum I'd like to see these be made easier to read, and using colors more colorblind-friendly than red/green. It would be great if the model scientists would put time and thought into making more explanatory ones, because not only will it benefit people reading the User Guide, but we'll be able to use them in training presentations later (so please provide the original graphic-design version too (PPT or whatever), not just the final png). I could think about it more and/or work with them on that if it's useful, but I'd have to understand the model way better than I currently do before making suggestions.

Without having a project to work through the outputs with, I also don't have an immediate suggestion for which intermediate files should be moved into the output folder, but could give it more thought if it's useful. I'd think that the model scientists would have a better idea of what they want to emphasize. Perhaps when they write explanations of the outputs and how to use them, this will become clearer. (Yes, I'm putting things on the scientists' plate, as I feel that these are critical parts of model design that should be standardly produced). It also seems like something easy to change later, and not critical to get "right" off the bat.

And thank you for all of the other additions and adjustments that you made!

phargogh:
Thanks for your feedback @newtpatrol !  I'll include your requests in the upcoming (very soon) batch of requests for elaboration from the science team.

phargogh:
I think this is as good as it's going to get for the release.  Merging.

PR #111, opened 2023-02-17
Chinese translation

emlys:
Adds the Chinese RST files. Since we did not get translated images, they point to the original English images.


dcdenu4 (approved):
Woohoo!!! Thanks @emlys for doing the leg work on this.

dcdenu4:
Hey @emlys , I had one question I just remembered. Were we going to update the Citation and Contributions section in this PR?

dcdenu4:
Hey @Emily, thanks for updating the Chinese UG contributor list. I noticed that on that commit that the dev build of the UG was not uploaded as an archive, so I couldn't see the latest via google cloud storage.... Should it have uploaded though?

I also noticed the citation was left the same, I can't remember if we decided that should wait for a proper translation or not.

Thanks!

emlys:
> the dev build of the UG was not uploaded as an archive

Good catch! Looks like it's failing to upload because the files already exist, and overwriting them requires delete access to the bucket. I'll make a new service account with delete access.

I didn't realize the citation had changed too, but I think we can copy that over without translation.

dcdenu4 (approved):
Thanks @emlys !

PR #117, opened 2023-02-28
remove readthedocs config

emlys:
Remove the config for our ReadTheDocs build. When we're sure we want to turn this off, I'll also delete the user's guide project on readthedocs.org.


dcdenu4 (approved):
Thanks @emlys . I think we can go ahead and turn this off from the readthedocs side.

PR #122, opened 2023-03-22
add scrollbar to sidebar on overflow.

davemfish:
Some table-of-contents (SDR) are very long and the content was not fully visible with a `fixed_sidebar`. Now the sidebar will have its own scrollbar when the content is too long.


emlys (approved):
Makes sense to me! Not sure why the requirements was pointing to the UNA branch, but we probably don't need to do that any more.

PR #123, opened 2023-05-02
Add subwatershed delineation `investspec` parameters

phargogh:
This is the UG counterpart to https://github.com/natcap/invest/pull/1291 for exposing subwatershed delineation to the user.

### DIFF
--- a/source/en/routedem.rst
+++ b/source/en/routedem.rst
@@ -28,6 +28,10 @@ The freshwater models in InVEST are routed across a digital elevation model (DEM
 
 * Create streams based on flow accumulation.
 
+* Determine the Strahler Stream Order of streams.
+
+* Delineate subwatersheds based on streams.
+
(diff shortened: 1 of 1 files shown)


source/en/routedem.rst
```diff
@@ -28,6 +28,10 @@ The freshwater models in InVEST are routed across a digital elevation model (DEM
 
 * Create streams based on flow accumulation.
 
+* Determine the Strahler Stream Order of streams.
```
dcdenu4, line 31: I mentioned this over in the InVEST PR too but would it be worth elaborating on what Strahler Stream order is? I don't think it's mentioned at all in the UG. Or is it common knowledge enough for us not to have to have an explainer?

dcdenu4 (commented):
Thanks @phargogh, just a quick comment / question!

dcdenu4:
Oh, and should this go into a bugfix release branch? Since these changes won't go into effect until the next InVEST release?

source/en/routedem.rst
phargogh: Yeah, this is new in the InVEST context (not that we actually use Strahler orders in InVEST), but I think this is something that's pretty standard in hydrology.  It seems to me that if we need to explain Strahler stream orders, then we should probably also explain flow accumulation!  What do you think?

phargogh:
@dcdenu4 thanks for the feedback!  I just updated the target branch.

source/en/routedem.rst
dcdenu4: Let's punt on it! Feels more like a science communication thing and it'd be great if we could defer to other NatCappers for that.

PR #125, opened 2023-06-19
Updating the citation and build information.

phargogh:
This PR updates the suggested citation in response to [a recent discussion on slack](https://stanford-natcap.slack.com/archives/CECFUU1DW/p1686941898377019). 

Specific changes are:

* Citation year is now pulled from the year of the latest commit.
* Citation version is the last git-tagged release.
* Git build information is moved to the HTML page footer.
* Download page link added

The git build information still has all the detail it always did, it's just now hidden.

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -1,6 +1,8 @@
+import datetime
 import os
-import sys
 import subprocess
+import sys
+
 import setuptools_scm  # Just fail the process if this can't be found.
 
 # add to the path so that sphinx can find our custom extension
@@ -39,7 +41,6 @@ master_doc = 'index'
 
 # General information about the project.
 project = 'InVEST'
-copyright = 'The Natural Capital Project'
 
 # configure user-agent to avoid 403 errors on linkcheck
 # https://github.com/sphinx-doc/sphinx/issues/7369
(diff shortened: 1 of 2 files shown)


source/conf.py
```diff
@@ -122,3 +129,25 @@
...
+try:
+    commit_year = subprocess.run(
+        ['git', 'show', '-s', '--format=%cs', 'HEAD'],
+        capture_output=True).stdout.decode('ASCII').strip().split('-')[0]
+except subprocess.CalledProcessError:
+    # If there's an error, default to the current year.
+    commit_year = datetime.datetime.now().year
+
+# Here we expose variables for use in our RST pages.
+rst_prolog = f"""
```
dcdenu4, line 145: I'm not certain what `rst_prolog` is for, but I was wondering if `|version| replace:: {version}` should also be in this list, since we're exposing it to use in the RST? I could be interpreting this wrong.

dcdenu4 (changes requested):
Thanks @phargogh! One comment / suggestion, if I'm interpreting correctly.

source/conf.py
phargogh: It looks like `rst_prolog` is inserted at the beginning of every source file as it is being read, so it allows us to do these kinds of global substitutions. https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-rst_prolog

It does look like this is a very general bit of RST, though, so we could probably insert any text or RST we wanted to!

PR #132, opened 2023-09-07
Add actions job to report translation status

emlys:
natcap/invest#1403

I didn't find any simple way to get a metric like "percent changed" of the source RST, but hopefully this gives you some idea of the amount of changes.


dcdenu4 (changes requested):
Hey @emlys , I think we decided to punt on this until we thought of what would be most useful for Lisa and others, as far as generating an easily digestible update of what has changed.

PR #140, opened 2024-04-29
Update translations

emlys:


dcdenu4 (commented):
Thanks @emlys ! Is there anything in particular I should keep an eye out for reviewing that'd be helpful? I suspect anything wonky can be found and fixed afterwards.

And, since these are changes that are already present in the English version, seems to make sense to pull them right into `main`.

dcdenu4 (approved):
Looks good.

PR #152, opened 2025-01-08
Clarified impact of intensity rating

### LINKED ISSUE #150: Clarify spatial impact of stressor rating in HRA
A [recent user](https://community.naturalcapitalproject.org/t/hra-spatially-explicit-criteria-layers-not-factoring-correctly-into-risk-calculation-invest-3-14-2/5642/6) was confused about why a stressor's rating (in the E/C criteria table) was not affecting the entire buffer area (defined in the habitat stressor table). It could be helpful to add a note in the users guide to clarify that ratings only affect the pixels directly underneath the stressor and do not extend through the buffer zone. 

I propose adding a note like "The intensity rating assigned to a stressor in the Exposure/Consequence Criteria Scores table only affects the area directly beneath that stressor, i.e. the rating does not extend through the buffer area that the stressor may be assigned in the habitat stressor table" (perhaps under the paragraph beginning with "**Intensity rating.** Exposure depends not only on whether the habitat and stressor...").

claire-simpson:
Added a note to clarify that the intensity rating that a user assigns to a stressor will only impact the footprint of the stressor and will not affect the buffer area around that stressor

Fixes #150

### DIFF
--- a/source/en/habitat_risk_assessment.rst
+++ b/source/en/habitat_risk_assessment.rst
@@ -82,6 +82,7 @@ The risk of human activities to habitats or species is modeled in five steps.
 
 where :math:`E_{jkl}` is the exposure score specific to habitat *j*, from stressor *k* in location *l*; :math:`C_{jkl}` is the consequence score, :math:`e_{ijkl}` is the exposure rating criterion *i*, specific to habitat *j* and stressor *k* and location *l*; :math:`c_{ijkl}` is the consequence rating. :math:`d_{ijkl}` represents the data quality rating, :math:`w_{ijkl}` represents the importance weighing for criterion. *N* is the number of criteria evaluated for each habitat.
 
+
 **Step 2.** The second step combines the exposure and response values to produce a distance-weighted risk value for each stressor-habitat combination in each grid cell. There are two options for risk calculation and three options for distance-based weighting.
 
 For Euclidean Risk calculation, risk to habitat *j* caused by stressor *k* in each location (i.e. cell) *l* is calculated as the Euclidean distance from the origin in the exposure-consequence space, where average exposure (:eq:`exposure`) is on one axis and the average consequence score (:eq:`consequence`) is on the other.
@@ -315,9 +316,7 @@ Default Exposure Criteria
 
    *Choose "0" to exclude this criterion from your assessment.*
 
-      .. note:: The intensity rating applies only to the area directly impacted by the stressor and does not extend through the buffer zone that may be defined for that stressor.
-
-
+ 
 4. **Management strategy effectiveness rating.** Management can limit the negative impacts of human activities on habitats. For example, regulations that require a minimum height for overwater structures reduce the shading impacts of overwater structures on submerged aquatic vegetation. Thus, effective management strategies will reduce the exposure from stressors to habitats or species. The effectiveness of management of each stressor is scored relative to other stressors in the region. So if there is a stressor that is very well managed such that it imparts much less stress on the system than other stressors, classify management effectiveness as "very effective." In general, however, the management of most stressors is likely to be "not effective." After all, you are including them as stressors because they are having some impact on habitats. You can then use this criterion to explore changes in management between scenarios, such as the effect of changing development from high impact (which might receive a score of "not effective") to low impact (which might receive a score of "very effective)." As with all criteria, higher numbers represent greater exposure and result in higher risk scores.
 
    *If criteria are scored on a 1-3 scale, the following is a suggestion for scoring management effectiveness:*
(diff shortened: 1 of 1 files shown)


claire-simpson:
@phargogh There are multiple places that this note could be added so I'd appreciate feedback on whether this is the best location!

source/en/habitat_risk_assessment.rst
```diff
@@ -315,6 +315,9 @@ Default Exposure Criteria
 
    *Choose "0" to exclude this criterion from your assessment.*
 
+      .. note:: The intensity rating applies only to the area directly impacted by the stressor and does not extend through the buffer zone that may be defined for that stressor.
```
phargogh, line 318: I agree that a `.. note::` is a great way to present this information!

On the content of the suggestion, though, I'm pretty sure a scalar rating does impact the area impacted, including through the buffered area.  The confusion for the user was that they provided spatially explicit criteria scores, but only provided them for the geometry of the stressor, not for the whole area that the stressor affected.

So I think I have 2 suggestions here:
1. Might I suggest we move the `.. note::` to the section below titled "Using Spatially Explicit Criteria"?
2. I might update the text of the message to clarify what the user needs to provide, maybe something like the below?

> When providing spatially explicit criteria layers, each layer must have pixel values or features that, when rasterized, cover the entire area affected by the stressor.

phargogh (changes requested):
Yep, I totally agree that there are multiple places where this information could be displayed (a downside of having a too-complicated chapter), and might I suggest we move this note to the spatially explicit criteria heading?

I also had a suggestion about the nature of the note, since I _think_ (but could be wrong about this!) that a scalar rating does apply throughout the whole buffer zone, while a spatially explicit rating covers only the area it covers when rasterized.  If this is true, moving the comment to the spatially explicit criteria area and also clarifying the coverage the rating layer must have might help clear up the original confusion.

Curious what you think, and I may well be wrong about the above!

source/en/habitat_risk_assessment.rst
claire-simpson: Great points! I don't think I fully realized that the rating field was applied differently to the buffer zone depending on whether the rating was constant or spatially explicit. 

>1. Might I suggest we move the .. note:: to the section below titled "Using Spatially Explicit Criteria"?

I agree that adding the note in the spatially explicit criteria section is a better option. 

>When providing spatially explicit criteria layers, each layer must have pixel values or features that, when rasterized, cover the entire area affected by the stressor.

I am slightly concerned that this revised comment isn't explicit enough about how the buffer is utilized. What about combining the notes to have it be:

> When providing spatially explicit criteria layers, each layer must have pixel values or features that, when rasterized, cover the entire area affected by the stressor. The spatially explicit intensity rating will not extend through the buffer zone that may be defined for that stressor.

phargogh: >the rating field was applied differently to the buffer zone depending on whether the rating was constant or spatially explicit

>The spatially explicit intensity rating will not extend through the buffer zone that may be defined for that stressor.

My understanding is that the rating calculation uses pixel values that represent criteria scores.  If a buffer is provided, then the criteria criteria score applies through that buffer, modified by a decay equation if selected.  But if a user provides a spatially explicit criteria score, then we just use the raster version of the criteria layer and buffering isn't a part of the criteria score at all.  The implication of this is that if there's no score for part of the area that is buffered, then that will have unexpected results as this user experienced.

So for this part:

> The spatially explicit intensity rating will not extend through the buffer zone that may be defined for that stressor.

I'm pretty sure the spatially explicit rating will cover whatever areas the user has it cover ... which might mean that it extends throughout the buffer zone!

claire-simpson: >I'm pretty sure the spatially explicit rating will cover whatever areas the user has it cover ... which might mean that it extends throughout the buffer zone! I don't think we can say that in all cases the intensity rating will not extend through the buffer zone, because it depends on what the user provides.

I guess my desire is to clarify that a user who adds a criterion with spatially explicit rating should not expect that rating to *automatically* extend through the buffer area. I.e., the user would need to manually buffer that area and add this buffered file into the Criteria Scores table. To me its not necessarily intuitive that a rating *would* be automatically extended throughout the buffer area if the user provides a constant rating in the Criteria Scores table, but the rating *would not* extend through the buffer area if the user provides a spatially explicit criterion for rating. Am I interpreting this correctly?

phargogh: Yes, I believe you are interpreting this correctly!  The assumptions that you describe are also really helpful, so thank you for clarifying where you have been coming from.

What if we made 2 distinct edits, then?

1. We could update the body of the model (under Step 1 of [Cumulative Risk to Habitats or Species from Multiple Stressors](https://storage.googleapis.com/releases.naturalcapitalproject.org/invest-userguide/latest/en/habitat_risk_assessment.html#cumulative-risk-to-habitats-or-species-from-multiple-stressors)) to better describe how scalar ratings behave (as a constant value) vs how spatially explicit ratings behave (which we really don't yet do very well, if at all)
2. We could update this section under spatially explicit criteria to reflect how spatially explicit criteria should be formatted, that there should be data for each pixel throughout the buffer zone, and that if there are missing data for a pixel within the buffer zone, that the criterion will be ignored on that pixel (it's effectively a criterion of 0)

Maybe that would better clarify things?

phargogh: I think this looks good!  Thanks for finding where to add this info.  I agree, it's a little awkward, and it doesn't help that this is the _longest_ chapter in the UG, despite the relative simplicity of the model.  Anyways, I'll approve and merge.  Thanks!

phargogh (approved):

source/en/habitat_risk_assessment.rst
claire-simpson: @phargogh Is it possible to reopen this PR? I had a realization about my changes, and I think I actually was a bit misleading in that the buffer areas can still be affected by a stressor, however the effect of the stressor will not be influenced by the spatially explicit rating unless the rating is explicitly specified throughout the buffer zone.. To that end, I am inclined to delete the last sentence I added in each of the two sections. Thoughts?

phargogh: I think maybe since you had created the PR, maybe you'd be able to reopen it?  If not, you could always make the change on the same branch and open a new PR on the same branch.

Personally, I think the two sentences are probably fine as-is, but please feel free to edit/update them for clarity!

PR #156, opened 2025-01-22
Removed misleading HRA text

claire-simpson:
The previous update to the HRA text was misleading with respect to the stressor buffers (i.e., implying that a buffer would not be impacted by a stressor unless a spatially explicit rating layer exists). I have removed the misleading text as a buffer area can still be affected by a stressor, but the spatially explicit _rating_ only comes into play for the pixels which the spatially explicit criterion overlaps. 

Relates to #150


phargogh (approved):

phargogh:
Thanks @claire-simpson !

PR #157, opened 2025-02-06
Add Metadata section to the Getting Started page.

davemfish:


source/en/getting_started.rst
```diff
@@ -246,6 +246,14 @@ To begin:
...
 
+Metadata
+========
+Results of InVEST models also include **metadata** documents that describe each dataset in the output workspace. These are the *".yml"*, or YAML, sidecar files in the output workspace. Each metadata file has the same filename as the dataset it describes, plus the *".yml"* extension. Open a YAML file in a text editor to read the metadata. It includes much of the same information as in the *Interpreting Results* section of the model chapter. This includes descriptions of fields in tables, of the bands in a raster, and other useful information. If you plan on sharing your InVEST results with others, we strongly encourage you to share the metadata documents along with your results.
```
claire-simpson, line 251: Similar to my comment on the other PR - maybe this section could be a good place to note that users can edit their metadata or link them to instructions for this purpose? I realize this is in the Getting Started section so maybe not much detail needed here

claire-simpson: More of a general question - but does the metadata created by geometamaker conform to a specific standard?

claire-simpson (approved):
Looks good!

source/en/getting_started.rst
davemfish: I think it would be good to have a link to more documentation on the metadata. For now the github link to geometamaker is the best source for that, which we're linking at the end of this section. If we improve the documentation there, or make a separate page, we can include the link here later. 

For InVEST metadata specifically, I'll add a note that the docs are editable, but I don't think most invest users will need to edit them because they already include lots of good info from the model itself.

claire-simpson (approved):

PR #166, opened 2025-03-15
update routing models to include D8 option

emlys:
Updating the user guides for NDR, SDR, and SWY to mention both D8 and MFD.

The checks will keep failing because the invest PR hasn't been merged yet, but it should be ok to ignore.


newtpatrol (commented):
These updates look great @emlys. My one suggestion would be that we add a new section to the User Guide providing a little more information about D8 and MFD, and the differences between them, including a couple of images showing how different the stream networks look. 

Since this is relevant to 3 models, perhaps it would be most efficiently included in the Data Sources section? I'm not sure if as many users would see it there versus earlier in the main text (like Defined Area of Outputs is), and don't feel strongly about that placement, but I think it would be helpful to include somewhere. I can help put that together (or not) as you please.

emlys:
Thanks @newtpatrol! I added a section to the Data Sources page and linked to it from the relevant models. If you would like to add anything or expand on when/why to choose one or the other, that would be great too.

newtpatrol (approved):
Super useful @emlys , thanks for writing that up. I see a few things that I might add/edit (like a blurb about considerations for choosing one algorithm over the other), but should I do that after it's merged?  Two of those things are worth noting here:

- The D8 description says that water flows to the left with a flowdir value of 4. Should that actually be to the right, from pixel X's point of view? 
- The comparison graphic would benefit from each image being labeled, so people don't need to read the caption to see what's going on. Just short labels like "MFD flow direction", "D8 flow direction", "MFD streams", "D8 streams".

newtpatrol:
> I'm not sure I'm understanding what you mean by pixel X's point of view. When looking at the raster grid, the neighbor represented by "4" is to the left of the pixel.

Sorry, I'm overthinking it, from the perspective of standing on that pixel in the landscape. Thinking about it more, I've confused myself, and your description is the most intuitive.

> This is unfortunately harder than it sounds. I'd like to avoid making the caption part of the image, because it's difficult to translate, and not machine-readable or searchable. Sphinx does not support having a grid of images that each have their own caption: https://github.com/sphinx-doc/sphinx/issues/4003. The four images would have to be stacked vertically, which doesn't look very good IMO. What do you think?

Wow, that's way more to consider than I realized. I was picturing it as a single PNG that had already combined the 4 maps, so Sphinx wouldn't need to format anything special. Translation is something I had not considered, but I'm not so concerned about the text in a graphic being searchable. 

It seems like an unfortunate tradeoff, making it take significantly more work for users to interpret the graphic, in exchange for translatability. When making graphics for papers and reports we always try to make them understandable with just a skim, not needing to read the whole (often lengthy) caption. I don't know enough about how either Sphinx or the translation process works to come up with a quick solution, but feel like it would be worth thinking about, so we don't end up with a whole document full of graphics with no legends or other descriptive text. As it is, I was also going to suggest a legend in each map, as that would illustrate your description of the flow dir values, but that legend probably would also need to be translated. 

So gow are other graphics in the UG being translated? Could we do something like have a PowerPoint version of each graphic, with anything that needs to be translated having an editable text box, and a new PNG generated from it? Is that too much work? I did the UNA graphics like this in PPT, and feel like we should be doing this anyway for all graphics, so that it's easier for future us to update them.

newtpatrol:
PS: I'm kind of crunched right now, and know that the InVEST release is pending, so I'll make additions to this UG section after everything gets merged.

emlys:
Other graphics in the UG often have both text in the image, and a caption that's plain text below the image. PPTX were not available for most of them, so our Spanish translator manually edited them somehow (like adding a text box covering the old text). That's a lot of work and I'm honestly impressed he was willing to do it. Images were not translated to Chinese. 

We could just shorten the caption to something like "D8 flow direction (top-left); MFD flow direction (top-right); D8 streams (bottom-left); MFD streams (bottom-right)".

Or here's another option: two figures with two images each and shorter captions.
image

I unfortunately don't have a ton of time to sink into this either, so whatever's most simple and effective. I guess the goal of the figure is just to communicate that D8 and MFD options are available; they generally produce similar results; but the exact shape of the streams will vary.

newtpatrol:
The option with two figures and short captions, plus the longer caption underneath looks good to me, it lets people interpret the maps quickly. Thanks for doing that. 

For later, let's discuss the idea of standardizing UG graphics so source is provided that is easily editable. Might not help our old ones, but it will improve things going forward.

jagoldstein:
Except, don't we want to include a legend, especially for Figure 4 (flow direction)?

emlys:
I'm not sure a legend would add anything for MFD flow direction - the values are very large 32-bit integers that are totally meaningless until they're decoded. Maybe it's too much information to include these images at all? The point is just to communicate what's said in the caption, but maybe that's enough on its own

emlys:
Thanks @newtpatrol and @jagoldstein, I am going to go ahead and merge this for the sake of the upcoming release, and we can continue to improve on it if needed

PR #172, opened 2025-03-28
Addition of Data Hub links to data sources sections

empavia:
Data Hub links were added to various locations in the User's Guide such as Data Sources, Annual Water Yield, SDR, and Seasonal Water Yield. 

Data links link to the search terms instead of the dataset itself to ensure URL stability if datasets are updated. For example, the DEM section links to the DEM search page on the catalog.

### DIFF
--- a/source/en/sdr.rst
+++ b/source/en/sdr.rst
@@ -617,7 +617,7 @@ In the United States, national maps of the erosivity index can be found through 
 The EPA has created a digital map that is available at https://archive.epa.gov/esd/archive-nerl-esd1/web/html/wemap_mm_sl_rusle_r_qt.html. The map is in a shapefile format that needs to be converted to raster, along with an adjustment in units.
 
 
-You can find InVEST-ready erodibility data on the NatCap Data Hub here: https://data.naturalcapitalproject.stanford.edu/dataset/?_tags_limit=0&tags=EROSIVITY
+You can find InVEST-ready rainfall erosivity (R factor) data on the NatCap Data Hub here: https://data.naturalcapitalproject.stanford.edu/dataset/?_tags_limit=0&tags=EROSIVITY
 
 
 Soil Erodibility (K)


source/en/data_sources.rst
```diff
@@ -39,6 +41,8 @@ Global land use data is available from several sources including:
 
 Data for the U.S. is provided by the USGS and Department of the Interior via the National Land Cover Database: https://www.usgs.gov/centers/eros/science/national-land-cover-database
 
+You can also find InVEST-ready LULC data on the NatCap Data Hub: https://data.naturalcapitalproject.stanford.edu/dataset/?_tags_limit=0&tags=LULC
```
newtpatrol, line 44: I'd just add "global", as in 
"You can also find InVEST-ready _**global**_ LULC data on the NatCap Data Hub..."
Unless anyone thinks we'll get into the business of providing sub-global LULCs, which I doubt (at least any time soon).

source/en/sdr.rst
```diff
@@ -616,6 +616,10 @@ In the United States, national maps of the erosivity index can be found through
 
 The EPA has created a digital map that is available at https://archive.epa.gov/esd/archive-nerl-esd1/web/html/wemap_mm_sl_rusle_r_qt.html. The map is in a shapefile format that needs to be converted to raster, along with an adjustment in units.
 
+
+You can find InVEST-ready erodibility data on the NatCap Data Hub here: https://data.naturalcapitalproject.stanford.edu/dataset/?_tags_limit=0&tags=EROSIVITY
```
newtpatrol, line 620: This should read "You can find InVEST-ready _**erosivity**_ data..." (not "erodibility").

newtpatrol (changes requested):
Just a couple of edits/suggestions, only one that's fixing an error, otherwise it looks good to me.

source/en/data_sources.rst
empavia: I am actually in the process of getting some local LULC rasters right now that are further processed versions of global layers! We will likely begin to host more local data as projects get published, so I figured keeping this broad would be best to start with.

davemfish (approved):

PR #176, opened 2025-05-08
Forest Carbon: Note 500km cap on global regression models

megannissel:
The Forest Carbon model only uses those global regression models that fall within 500km of the study area. If the AOI were to span the entire pantropics, every model could in theory be used, but generally this won't be the case. Since the models are weighted by distance, anyway, this shouldn't matter in practice -- but should still be noted in the user's guide.

Also, the number of features in the global regression models vector is 2313, not 2635.


phargogh (approved):
Looks good to me, thanks @megannissel !

PR #178, opened 2025-05-09
Wind Energy: Update model outputs

megannissel:
Related to [InVEST PR #1898](https://github.com/natcap/invest/pull/1898)

We're changing the outputs of the Wind Energy model. `wind_energy_points.shp` will now be the primary output, with several new fields. The old raster outputs will now serve only as intermediate outputs, and Density values won't be rasterized at all. Updating the User's Guide to reflect these changes.

### DIFF
--- a/source/en/wind_energy.rst
+++ b/source/en/wind_energy.rst
@@ -371,11 +371,26 @@ Valuation
 Interpreting Results
 ====================
 
-All output resolutions are based on the resolution of the supplied digital elevation model raster. When the resolution of the DEM exceeds the resolution of the wind data layers, pixel values are determined by using bilinear interpolation.
+Model Outputs
+-------------
+
+Output Folder
+^^^^^^^^^^^^^
+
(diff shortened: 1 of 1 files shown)


rmgriffin:
Looks good! Just one question on the writeup:
- DistVal is "included only if distance parameters were provided." Why isn't it provided if these are not given? Does the model skip the distance calculation step and so they aren't available?

dcdenu4:
>In which case an AOI and land polygon should generally be required inputs for all model runs.

We would love this to be the standard! Maybe we can follow up over the summer about this and other changes / features? But, I'd suggest moving forward with what we currently have that addressed the original issues and avoid scope creep for now.

dcdenu4:
>For foundation-based wind farms, the distance/depth constraint should apply to all outputs, as they are all conditional on actually being able to build a farm there.

I think it's also important to remember, and maybe this is something we lost sight of a little bit, is that the wind data points are not necessarily potential farm siting locations. Now that we've removed interpolated raster outputs we are kind of treating them as such, but the point outputs don't directly represent where you'd want to place a turbine farm.

dcdenu4 (approved):
Thanks @megannissel !

rmgriffin:
Yeah, might be worth flagging in the UG that issue about the points. Ultimately this is all handholding type stuff, if you understand the model well enough you should be able to parse the results with all of this in mind.

PR #192, opened 2025-07-01
Add freshwater calibration section

newtpatrol:


source/en/calibration_freshwater.rst
```diff
@@ -0,0 +1,150 @@
...
+
+Once the watershed is correctly generated, a GIS tool like Zonal Statistics is used to sum the relevant model result raster (such as sediment export or nutrient export) within the watershed. This summarized value is then compared with the observed data value. Alternatively, you can use the generated watershed as an input to the model, which will do the summarizing for you, creating a vector layer whose table contains the summarized values.
+
+Setting up the InVEST Python scripting environment
+==================================================
```
davemfish, line 88: Do you think it makes sense for this section to refer to our other docs on installing `natcap.invest` and running python scripts?
* https://invest.readthedocs.io/en/latest/installing.html
* https://invest.readthedocs.io/en/latest/scripting.html#setting-up-your-python-environment

The benefit is that we only have to keep this stuff updated in one place.

source/en/calibration_freshwater.rst
```diff
@@ -0,0 +1,150 @@
...
+	+ Install the InVEST Python package:
+		* ``conda install -c conda-forge natcap.invest``
+		* This command should install all dependencies because we’re using conda
+3. InVEST Python package version check
+	+ Launch python shell ``($ python)``:
+		* ``>>> import natcap.invest``
+		* ``>>> print(natcap.invest.__version__)``
+		* ``>>> exit()``
+	+ In this example, this installs InVEST v3.14.1, so update with pip:
+		* ``pip install natcap.invest==3.14.2``
```
davemfish, line 107: `conda install -c conda-forge natcap.invest` will always install the latest version, so there should not be a need to update right away. The suggestion to update the package using `pip` instead of `conda` kinda makes sense, but only if you know how those package managers differ and know that all your other dependencies don't need updating. I would probably just remove this `+` bullet point.

source/en/calibration_freshwater/SDR_script_example_manual_repeat.py
```diff
@@ -0,0 +1,55 @@
...
+    'lulc_path': 'G:\\Shared drives\\GEF_GreenFin\\Pilot_Countries\\Philippines\\DENR_Stanford_sharing\\SDR_Luzon_example_inputs\\LULC_2020_Luzon.tif',
+    'results_suffix': 'Luzon_example_kb1_8', # results filenames adjusted to reflect parameter change from baseline value
+    'sdr_max': 0.8,
+    'threshold_flow_accumulation': 1000,
+    'watersheds_path': 'G:\\Shared drives\\GEF_GreenFin\\Pilot_Countries\\Philippines\\DENR_Stanford_sharing\\SDR_Luzon_example_inputs\\Luzon_AOI.shp',
+    'workspace_dir': 'G:\\Shared drives\\GEF_GreenFin\\Pilot_Countries\\Philippines\\DENR_Stanford_sharing\\Calibration\\SDR_results', # Filepaths to each model input for adjusted parameter run
+} # UPDATE these filepaths to reflect the absolute paths on your computer
+
+if __name__ == '__main__':
+    natcap.invest.sdr.sdr.execute(args) # Run InVEST SDR model
```
davemfish, line 55: This script is pretty similar to the next one, so I might suggest only including that one and suggesting people save a separate script for each scenario. Or if you think it's useful to run multiple models in one script like this, I would suggest the following: Two separate arguments dictionaries (e.g. `args_scenario1` and `args_scenario2`). And then a single,

```
if __name__ == '__main__':
    natcap.invest.sdr.sdr.execute(args_scenario1)
    natcap.invest.sdr.sdr.execute(args_scenario2)
```

source/en/calibration_freshwater.rst
```diff
@@ -0,0 +1,150 @@
...
+	+ ``cd PATH``
+5. Run script:
+	+ ``python FILENAME.py``
+
+Example scripts
+===============
+
+InVEST Workbench can create a basic python script for any model. The script imports the necessary libraries, sets up logging, defines the model’s inputs and calls the model with those inputs. It is useful to start with this script, and build on it to create your sensitivity/calibration workflow. To create the script, within Workbench, go to an open model tab, select **Save as…** in the left toolbar, then select **Python script**. 
+
++ `Basic script for SDR <./calibration_freshwater/basic_SDR_script_example.py>`_, as generated by InVEST workbench.
```
davemfish, line 118: Hmm these links don't work; the python files are not included in the build. I think there are couple options,

1. Put these scripts somewhere else and use a complete URL (https://...) to reference them. They could be included in this repo, but the URL for them won't exist until after this PR is merged. So putting them in a another repo or a gist is also a good option. https://gist.github.com/. 
2. Use some other sphinx directive that will trigger these files to be included in the build. The only one that might work is `:download:` because python files will not be viewable in the browser. https://sphinx-intro-tutorial.readthedocs.io/en/latest/sphinx_roles.html#downloadable-files

This is a great use-case for a gist, I think. Then we could also easily add these to our cookbook as well. https://github.com/natcap/cookbook

davemfish (changes requested):
Looks great so far @newtpatrol . I have a couple suggestions though.

source/en/calibration_freshwater.rst
newtpatrol: Yes, great catch Dave. This section was copied from a quick HowTo that we wrote a while back, and I forgot about the readthedocs docs. I've updated that section to point to those two resources.

source/en/calibration_freshwater.rst
davemfish: I think the difference is that when the images are referenced, they use a sphinx directive like this `figure` directive:
```
.. figure:: ./annual_water_yield/watercycle.png
```
Then, during the build process, external files like this that are referenced are moved into other folders, like `_images`, which are included in the final package. So the placement of the file in this repo doesn't really matter, it's the "directive" that tells sphinx that the file needs to be packaged with the rest of the html. If you download and extract the zip file artifact you can see that the file and folder structure is not exactly the same as this repository. And the `.py` files are not present anywhere.

source/en/calibration_freshwater/SDR_script_example_manual_repeat.py
jagoldstein: Good points, I agree with Dave here, either suggest separate scripts for each scenario or to use different argument dictionaries. I lean towards the latter as I think fewer scripts would be simpler.

jagoldstein (approved):
This will be a much welcomed section in the UG! Thanks for all of your work on this @newtpatrol . My only comment was in agreeing with Dave about the "manual repeat" example script. That would probably be better with separate arg dictionaries for each scenario. I do think there's value in having multiple runs in a single script, but suggesting separate scripts would be fine too.

source/en/calibration_freshwater/SDR_script_example_manual_repeat.py
newtpatrol: It is, I was also wondering about whether to keep it, deleted its reference from the chapter.

source/en/calibration_freshwater.rst
newtpatrol: @davemfish I turned the 3 scripts that are currently referenced into public gist versions, updated the links in the calibration chapter, and they seem to be accessible in the current build. Let me know if they work for you, and if any other changes should be made to them.

davemfish (approved):
Thanks for the updates @newtpatrol , this looks great!

PR #196, opened 2025-07-28
Post dev build URLs to Slack for easy previewing

### LINKED ISSUE #195: Deploy dev builds for easy previewing
It would be convenient to deploy builds from other forks to some sort of dev-builds public bucket. That would create a public URL where the build can be viewed. Without that, previewing changes before they go into `main` means either building locally, or downloading and extracting the build artifact from github actions. If Actions could generate a direct link to a public URL, that would be quite nice.

davemfish:
Fixes #195 

GHA workflows will post to our #sw-invest-ug Slack channel after deploying an html build to our dev-builds bucket.

Forks will need to add a `SLACK_SW_INVEST_UG_WEBHOOK` repo secret in order to complete the new GHA step.

### DIFF
--- a/.github/workflows/post_url_to_slack.py
+++ b/.github/workflows/post_url_to_slack.py
@@ -13,8 +13,10 @@ def main(dev_build_url):
     if not dev_build_url:
         raise ValueError("The development build URL must be provided.")
 
+    # Webhooks for the SWTeam Notifier Slack App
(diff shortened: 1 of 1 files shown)


phargogh (approved):
Nice and simple, looks great to me!  Thanks @davemfish !

PR #198, opened 2025-08-25
SWY: Update inputs for Chinese and Spanish translations

megannissel:
In addition to updating the `et0_dir` and `precip_dir` keys on the English version of the UG, we also need to update the Chinese and Spanish translations to point to the new `et0_raster_table` and `precip_raster_table` inputs.


emlys (approved):
Thanks, and sorry for not catching that before!

PR #216, opened 2026-01-27
update build process to parse new version string format

### LINKED ISSUE #215: BUG: version string parsing in sphinx build
This was never brought up to speed when we switched from using `.post`-formatted versions to `.dev`.
https://github.com/natcap/invest.users-guide/blob/c3b2e53c19168a56dc7fb06c4cd0b171be3bc35e/source/conf.py#L57

It still works when building only in the context of this repo. But when building in the context of the invest repo, it gets a version string that can't be associated with a specific tag using `git tag --points-at '3.17.3.dev97+g5081c4005'` because the `3.17.3` tag (and release) does not exist yet.

However, userguide builds in invest GHA workflows are still working. And it may simply be working by coincidence:

```
Traceback (most recent call last):
  File "D:\a\invest\invest\setup.py", line 6, in <module>
    from Cython.Build import cythonize
ModuleNotFoundError: No module named 'Cython'
```
https://github.com/natcap/invest/actions/runs/21368912963/job/61519680788

This error when running `python setup.py --version` may simply cause the program to divert into the `except` clause and treat the build process as if it was happening in the userguide repo instead of the invest repo.

davemfish:
Fixes #215 

@phargogh I feel these changes might undo things you did for good reason, so I'm happy to revise these changes and find another way to the most recent prior tag. Thanks for taking a look.

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -74,11 +74,25 @@ except subprocess.CalledProcessError:
         git_version = git_version.replace('+', '+ug.')
 
 print(f"Version from git: {git_version}")
+
 # We decided on slack 2023-06-19 that the UG version in the citation should
-# only be the version in the latest
+# only be the version in the latest release
+# This tells us the latest release prior to HEAD
 version = subprocess.check_output(
-    ['git', 'describe', '--abbrev=0', '--tags']
-).decode('ASCII')
+    ['git', 'describe', '--abbrev=0', '--tags']).decode('ASCII').split('\n')[0]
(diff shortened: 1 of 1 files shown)


dcdenu4:
Sorry for jumping in.

>Is it OK if we have 2 tags pointing to the same revision and returning the earliest tag instead of the latest?

I think you make the argument well James, that we don't want the User Guide to appear "behind" or out of date even if it isn't.

PR #218, opened 2026-02-12
Update SDR equation

emlys:
For natcap/invest#2379


newtpatrol (approved):
This looks fine to me @emlys . The only thing i might add is here:
"The value of 1 is intuitively reasonable because it represents 100% sediment delivery - no sediment is trapped as it flows from one stream pixel to another."
To be very explicit, i'd add something like this:
"...since model calculations are not done within the stream network, we assume that no sediment is trapped as it flows from one stream pixel to another."

PR #239, opened 2026-06-12
build natcap.invest from source

davemfish:
This PR includes all the changes on the `release/3.20.0` branch. It also updates the build workflow to deal with building invest from source, since the 3.20.0 release does not include linux wheels.


emlys (approved):
Thanks @davemfish. I think the "upload to release bucket" step is the only part that won't run either in the PR or on your branch until after it's merged. So that is not being tested, but it's not a big deal if something breaks there.

PR #240, opened 2026-06-16
Clean up image files

emilyanndavis:
While looking through the diff in preparation for the next round of translation, I noticed the Workbench screenshots in the "Getting Started" section had been updated, but the old screenshots hadn't been removed. Digging a bit deeper, I found a couple of images of tables that aren't being used (because they've been replaced with inline tables generated from CSVs), as well as some duplicate and otherwise unused slides in the PPT source files.

I figured now is a good time to clean up a bit so we don't mistakenly ask someone to translate images we aren't using anymore.

Specifically, this PR does the following:
- removes outdated Workbench screenshot PNGs,
- removes outdated Workbench screenshots from the PPT source,
- renames current Workbench screenshot PNGs,
- updates image references in getting_started.rst,
- removes a no-longer-needed temp file,
- removes unused images of tables (NDR_biophysical_table_3-16 and SWY_climate_zone_table, both of which have been replaced with CSVs),
- removes duplicate/unused slides from Carbon and Blue Carbon PPT files,
- renames Carbon and Blue Carbon PPT files so they're easier to associate with those models, and
- adds PPT source for the UMH flowchart.


davemfish (approved):
Looks good to me, thanks for cleaning this up.

PR #248, opened 2026-08-20
Specify Wind Energy AOI projection requirements

### LINKED ISSUE #246: Wind Energy model Data Needs section does not specify that AOI must be projected in units meters
The `MODEL_SPEC` _does_ specify that the AOI must have these projection requirements, and we are theoretically pulling this info for the UG from the `MODEL_SPEC`

megannissel:
Fixes #246 

While the baseline file type, units, and description are pulled directly from the model spec, things like projection requirements are not. As such, I've updated the Wind Energy AOI description to include its projection requirements.

### DIFF
--- a/source/en/wind_energy.rst
+++ b/source/en/wind_energy.rst
@@ -274,7 +274,7 @@ Data Needs
   - :investspec:`wind_energy wind_data_path.columns.k`
   - :investspec:`wind_energy wind_data_path.columns.ref`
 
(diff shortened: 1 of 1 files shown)


claire-simpson (approved):
> While the baseline file type, units, and description are pulled directly from the model spec, things like projection requirements are not. As such, I've updated the Wind Energy AOI description to include its projection requirements.

Hmm, I wonder why we don't pull `projection` from the model spec.. I believe we've discussed greater integration between the UG and model spec in the past, so maybe this just hasn't happened yet? Anyway, this PR looks good as an immediate solution!

